PR #3, opened 2020-03-19
Merge type id and definition

ascjones:
Migrated from https://github.com/type-metadata/type-metadata/pull/42

- Renames `TypeId` to `Type` and merges in `TypeDef`
- Removes `HasTypeDef` trait etc.
- Introduces `Composite` (for structs) and `Variant` (for enums) for user defined types
- Introduce `Field` with optional name. Builders enforce construction of either *all* named (structs) or all unnamed (tuples). 
- Introduce `Variant` with optional fields. For "C-like" enums, a builder enforces that *all* variants have no fields and an optional discriminant. For enums with fields, we reuse the `Field` type and builders for "struct" and "tuple" variants.
- Standardizes json names (originally https://github.com/type-metadata/type-metadata/pull/36)

### Downstream PRs

- **ink!**: https://github.com/paritytech/ink/pull/296
- **polkadot-js**: https://github.com/polkadot-js/api/pull/2118

### todo

- [ ] Fix up documentation
- [x] Rename `TypeId` struct?
- [x] Test with `ink!`
- [x] Test with polkadotjs api

### comments from legacy PR

- [ ] Clarify Vec/Btree representation https://github.com/type-metadata/type-metadata/pull/42#discussion_r394522324
- [x] Decide on flattening `path` or not https://github.com/type-metadata/type-metadata/pull/42#discussion_r394487659 (**keep it flat**)
- [x] Confirm skipping serialization of empty fields works: https://github.com/type-metadata/type-metadata/pull/42#discussion_r394493884 (**yes it does**)
- [x] Keep comment width wrap formatting? https://github.com/type-metadata/type-metadata/pull/42#discussion_r393852061 (**removed**)


Robbepop:
I was having a chat with @jacogr and we had the idea to conceptualize certain common structures in the generated JSON SCALE info.
For example we conceptualize the structure of mapping types.
A mapping type encodes as a sequence over some other type that has a `key` and `value` field. That's it. Below you can see the JSON that could conceptualize such a scenario concretely. Note that it might be nice to have so-called Voldemord composites that do not have a name since it isn't needed here. We just generate the type at index 1 upon serialization and will never need its name.

```json
{
    names: [
        "BTreeMap", # ID: 1
        "KeyValue", # ID: 2
        "K",        # ID: 3,
        "V",        # ID: 4,
        "key",      # ID: 5,
        "value",    # ID: 6,
    ]
    types: [
        { # ID: 1
            composite: {
                path: [ 2 ], # We actually don't need this.
                             # Could we define composite types without names?
                             # Does that make sense at all? It would be similar
                             # to the concept of Voldemord types.
                fields: [
                    { name: 5, type: 4 },
                    { name: 6, type: 5 },
                ]
            }
        },
        { # ID: 2
            sequence: {
                type: 1
            }
        },
        { # ID: 3
            composite: {
                path: [ 1 ],
                params: [ 4, 5 ],
                fields: [
                    type: 2
                ]
            }
        },
        ...
    ],
}
```

src/impls.rs
```diff
@@ -113,200 +90,119 @@ impl_metadata_for_tuple!(A, B, C, D, E, F, G, H);
...
-		TypeIdCustom::new("Option", Namespace::prelude(), tuple_meta_type![T]).into()
+	fn type_info() -> Type {
+		TypeVariant::new("Option", Namespace::prelude())
+			.type_params(tuple_meta_type![T])
+			.variants(
+				Variants::with_fields()
+					.variant_unit("None")
+					.variant("Some", Fields::unnamed().field_of::<T>()),
+			)
+			.into()
```
Robbepop, line 114: Seeing this we still have the problem that our encoding is not generics-aware which it should imo. So we end up having a different type definition for each `Option<T>` with a different `T` instead of re-using the one generic definition. I have made some thoughts about this and came to the conclusion that it should even be fairly simple to actually implement this with some additional built-ins.

There are actually 2 ways we could implement this:

By introducing two new built-in type classes `generic_type` and `generic_param`:
```
{
    "strings": {
        "Option", # ID 1
        "Some",   # ID 2
        "None",   # ID 3
        "T",      # ID 4
        "Foo",    # ID 5
        "Bar",    # ID 6
        "Baz",    # ID 7
    }
    "types": {
        { # ID 1
            "variant": {
                "path" [ 1 ],
                "params": [ 4 ]
                "variants": [
                    {
                        "name": [ 2 ],
                        "type": [ 2 ],
                    },
                    {
                        "name": [ 3 ],
                    }
                ],
            }
        },
        { # ID 2
            # Generic parameter with name `"T"`.
            "generic_param": {
                "name": 4
            }
        },
        { # ID 3
            "primitive": "u32"
        },
        { # ID 4
            "primitive": "bool"
        },
        { # ID 5
            # Forwards to the `Option<u32>` type.
            "generic_type": {
                "type": 1,
                "params": 3,
            }
        },
        { # ID 6
            # Demonstrates usage of a type that contains a Option<u32> type.
            "composite": {
                "path": [ 5 ],
                "fields": [
                    {
                        "type": 5
                    }
                ],
            }
        },
        { # ID 7
            # Forwards to the `Option<bool>` type.
            "generic_type": {
                "type": 1,
                "params": 4,
            }
        },
        { # ID 8
            # Demonstrates usage of a type that contains a Option<bool> type.
            "composite": {
                "path": [ 6 ],
                "fields": [
                    {
                        "type": 7
                    }
                ],
            }
        },
        { # ID 9
            # Demonstrates usage of a non-generic type.
            "composite": {
                "path": [ 7 ],
                "field": [
                    {
                        "type": 3,
                    },
                    {
                        "type": 4,
                    },
                ]
            }
        }
    }
}
```

By introducing flags for all `type` fields:
```
{
    "strings": {
        "Option", # ID 1
        "Some",   # ID 2
        "None",   # ID 3
        "T",      # ID 4
        "Foo",    # ID 5
        "Bar",    # ID 6
        "Baz",    # ID 7
    }
    "types": {
        { # ID 1
            "variant": {
                "path" [ 1 ],
                "params": [ 4 ]
                "variants": [
                    {
                        "name": [ 2 ],
                        "type": [
                            {
                                # 4 indexes into the string table and must have
                                # the same index as one of the identifier in `params`.
                                "parameter": 4
                            }
                        ],
                    },
                    {
                        "name": [ 3 ],
                    }
                ],
            }
        },
        { # ID 3
            "primitive": "u32"
        },
        { # ID 4
            "primitive": "bool"
        },
        { # ID 5
            # Demonstrates usage of a type that contains a Option<u32> type.
            "composite": {
                "path": [ 5 ],
                "fields": [
                    {
                        # Demonstrates usage of the generic type at index 1 (`Option<T>`)
                        # with a parameter of 4 where 4 indexes into the type table as `u32`.
                        "type": {
                            "generic": {
                                "type": 1,
                                "params": 3,
                            },
                        }
                    }
                ],
            }
        },
        { # ID 6
            # Demonstrates usage of a type that contains a Option<bool> type.
            "composite": {
                "path": [ 6 ],
                "fields": [
                    {
                        "type": {
                            # Demonstrates usage of the generic type at index 1 (`Option<T>`)
                            # with a parameter of 4 where 4 indexes into the type table as `bool`.
                            "generic": {
                                "type": 1,
                                "params": 4,
                            },
                        }
                    }
                ],
            }
        },
        { # ID 7
            # Demonstrates usage of a non-generic type.
            "composite": {
                "path": [ 7 ],
                "field": [
                    {
                        "type": {
                            # 3 indexes into the type table and refers to a non
                            # generic (aka concrete) type, in this case `u32`.
                            "concrete": 3
                        },
                    },
                    {
                        "type": {
                            # 4 indexes into the type table and refers to a non
                            # generic (aka concrete) type, in this case `bool`.
                            "concrete": 4
                        }
                    },
                ]
            }
        }
    }
}
```

src/tests.rs
```diff
@@ -20,79 +20,92 @@ use core::marker::PhantomData;
...
 fn array_primitives() {
 	// array
-	assert_type_id!([bool; 3], TypeIdArray::new(3, bool::meta_type()));
+	assert_type!([bool; 3], TypeArray::new(3, bool::meta_type()));
 	// nested
-	assert_type_id!([[i32; 5]; 5], TypeIdArray::new(5, <[i32; 5]>::meta_type()));
+	assert_type!([[i32; 5]; 5], TypeArray::new(5, <[i32; 5]>::meta_type()));
 	// slice
-	assert_type_id!([bool], TypeIdSlice::new(bool::meta_type()));
+	assert_type!([bool], TypeSlice::new(bool::meta_type()));
```
Robbepop, line 106: Please rename to `TypeSequence` since that is the more general term for it.

src/ty/fields.rs
```diff
@@ -0,0 +1,173 @@
...
+pub enum NoFields {}
+/// A fields builder only allows named fields (e.g. a struct)
+pub enum NamedFields {}
+/// A fields builder only allows unnamed fields (e.g. a tuple)
+pub enum UnnamedFields {}
+
+pub struct Fields<T = NoFields> {
+	fields: Vec<Field<MetaForm>>,
+	marker: PhantomData<fn() -> T>,
+}
```
Robbepop, line 106: I think it makes sense here as well to split `Fields` from `FieldsBuilder` and have no generic parameters on the `Fields` type itself.

test_suite/tests/json.rs
```diff
@@ -55,30 +45,27 @@ fn test_unit_struct() {
...
-		"id": {
-			"custom.name": 1,
-			"custom.namespace": [2],
-			"custom.params": [],
-		},
-		"def": {
-			"tuple_struct.types": []
+		"composite": {
+			"name": 1,
+			"namespace": [2],
```
Robbepop, line 50: I think we should merge `name` and `namespace` into a non-empty array of string table indices called `path`.

derive/src/lib.rs
```diff
@@ -17,26 +17,204 @@
...
+				.variant_unit(#v_name)
+			},
+		}
+	});
+	quote! {
+		variants(
+			_scale_info::Variants::with_fields()
+				#( #variants)*
+		)
+	}
```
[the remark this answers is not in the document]
ascjones: I believe `synstructure` is geared towards deriving instance methods destructuring `&self` (correct me if I'm wrong)

src/impls.rs
ascjones: https://github.com/paritytech/scale-info/issues/6

src/impls.rs
```diff
@@ -113,200 +90,119 @@ impl_metadata_for_tuple!(A, B, C, D, E, F, G, H);
...
-impl<T> HasTypeId for Box<T>
-where
-	T: HasTypeId + ?Sized,
-{
-	fn type_id() -> TypeId {
-		T::type_id()
+	fn type_info() -> Type {
+		TypeComposite::new("BTreeMap", Namespace::prelude())
+			.type_params(tuple_meta_type![(K, V)])
+			.fields(Fields::named().field_of::<[(K, V)]>("elems"))
```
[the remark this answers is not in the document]
ascjones: https://github.com/paritytech/scale-info/issues/5

src/impls.rs
```diff
@@ -104,7 +104,8 @@ where
 	T: Metadata + 'static,
 {
 	fn type_info() -> Type {
-		TypeVariant::new("Option", Namespace::prelude())
+		TypeVariant::new()
+			.path(Path::new().ident("Option"))
```
Robbepop, line 108: I'd rather go for a special constructor so that users have to opt-in to create a Voldemord type which should be a rather rare thing instead of making users opt-in to create a common named type.

Robbepop: E.g. we could have 4 constructors
1. `Path::voldemord`
2. `Path::prelude` (with prelude namespace)
3. `Path::new(name, namespace)` (or so)

Robbepop: And maybe `Path::prelude` should actually be hidden from users.

src/tests.rs
```diff
@@ -47,12 +47,15 @@ fn primitives() {
 fn prelude_items() {
 	assert_type!(
 		String,
-		TypeComposite::new("String", Namespace::prelude()).fields(Fields::named().field_of::<Vec<u8>>("vec"))
+		TypeComposite::new()
+			.path(Path::new().ident("String"))
+			.fields(Fields::named().field_of::<Vec<u8>>("vec"))
```
Robbepop: For now I'd simply make `String` generate the same encoding as `str`.

README.md
```diff
@@ -1,49 +1,99 @@
...
+- **Primitives:** `bool`, `char`, `str`, `u8`, `u16`, `u32`, `u64`, `u128`, `i8`, `i16`, `i32`, `i64
+`, `i128`.
+
+- **Slice:** Variable size `&[T]` for any `T` which implements `TypeInfo`
+
+- **Array:** Fixed size `[T: $n]` for any `T` which implements `TypeInfo`, where `$n` is one of the
+ predefined sizes.
+ 
+- **Tuple:** Tuples consisting of up to 10 fields with types implementing `TypeInfo`.
+
```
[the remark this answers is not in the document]
ascjones: #8

src/ty/path.rs
```diff
@@ -63,16 +63,50 @@ impl IntoCompact for Path {
...
+		Self::from_segments(segments)
+	}
+
+	/// Create an empty path for types which shall not be named
+	///
+	/// # Note
+	///
+	/// Returns an always `Ok` Result to match the other constructor signatures
+	#[allow(unused)]
+	pub(crate) fn voldermort() -> Result<Path, PathError> {
```
Robbepop, line 78: `voldermort` -> `voldemort` pls correct

src/ty/path.rs
```diff
@@ -63,16 +63,50 @@ impl IntoCompact for Path {
...
+		Ok(Path { segments: Vec::new() })
 	}
 
+	/// Crate a Path for types in the Prelude namespace
+	///
+	/// # Errors
+	///
+	/// - If the supplied ident is an invalid Rust identifier
+	pub fn prelude(ident: <MetaForm as Form>::String) -> Result<Path, PathError> {
```
Robbepop, line 87: Why is this `pub`?

src/ty/path.rs
```diff
@@ -63,16 +63,50 @@ impl IntoCompact for Path {
 
 impl Path {
 	/// Start building a Path with PathBuilder
-	#[cfg_attr(feature = "cargo-clippy", allow(clippy::new_ret_no_self))]
-	pub fn new() -> PathBuilder {
-		PathBuilder::new()
+	pub fn new(module_path: <MetaForm as Form>::String, ident: <MetaForm as Form>::String) -> Result<Path, PathError>
```
Robbepop: I think it makes sense to switch parameters because `module_path` might be a longer expression in general.

PR #12, opened 2020-05-05
Generics

### LINKED ISSUE #6: Generics
As a follow up to #3, from @Robbepop: https://github.com/paritytech/scale-info/pull/3#discussion_r401521629. Should be included in a `0.1` release.

Seeing this we still have the problem that our encoding is not generics-aware which it should imo. So we end up having a different type definition for each `Option<T>` with a different `T` instead of re-using the one generic definition. I have made some thoughts about this and came to the conclusion that it should even be fairly simple to actually implement this with some additional built-ins.

There are actually 2 ways we could implement this:

By introducing two new built-in type classes `generic_type` and `generic_param`:
```
{
    "strings": {
        "Option", # ID 1
        "Some",   # ID 2
        "None",   # ID 3
        "T",      # ID 4
        "Foo",    # ID 5
        "Bar",    # ID 6
        "Baz",    # ID 7
    }
    "types": {
        { # ID 1
            "variant": {
                "path" [ 1 ],
                "params": [ 4 ]
                "variants": [
                    {
                        "name": [ 2 ],
                        "type": [ 2 ],
                    },
                    {
                        "name": [ 3 ],
                    }
                ],
            }
        },
        { # ID 2
            # Generic parameter with name `"T"`.
            "generic_param": {
                "name": 4
            }
        },
        { # ID 3
            "primitive": "u32"
        },
        { # ID 4
            "primitive": "bool"
        },
        { # ID 5
            # Forwards to the `Option<u32>` type.
            "generic_type": {
                "type": 1,
                "params": 3,
            }
        },
        { # ID 6
            # Demonstrates usage of a type that contains a Option<u32> type.
            "composite": {
                "path": [ 5 ],
                "fields": [
                    {
                        "type": 5
                    }
                ],
            }
        },
        { # ID 7
            # Forwards to the `Option<bool>` type.
            "generic_type": {
                "type": 1,
                "params": 4,
            }
        },
        { # ID 8
            # Demonstrates usage of a type that contains a Option<bool> type.
            "composite": {
                "path": [ 6 ],
                "fields": [
                    {
                        "type": 7
                    }
                ],
            }
        },
        { # ID 9
            # Demonstrates usage of a non-generic type.
            "composite": {
                "path": [ 7 ],
                "field": [
                    {
                        "type": 3,
                    },
                    {
                        "type": 4,
                    },
                ]
            }
        }
    }
}
```

By introducing flags for all `type` fields:
```
{
    "strings": {
        "Option", # ID 1
        "Some",   # ID 2
        "None",   # ID 3
        "T",      # ID 4
        "Foo",    # ID 5
        "Bar",    # ID 6
        "Baz",    # ID 7
    }
    "types": {
        { # ID 1
            "variant": {
                "path" [ 1 ],
                "params": [ 4 ]
                "variants": [
                    {
                        "name": [ 2 ],
                        "type": [
                            {
                                # 4 indexes into the string table and must have
                                # the same index as one of the identifier in `params`.
                                "parameter": 4
                            }
                        ],
                    },
                    {
                        "name": [ 3 ],
                    }
                ],
            }
        },
        { # ID 3
            "primitive": "u32"
        },
        { # ID 4
            "primitive": "bool"
        },
        { # ID 5
            # Demonstrates usage of a type that contains a Option<u32> type.
            "composite": {
                "path": [ 5 ],
                "fields": [
                    {
                        # Demonstrates usage of the generic type at index 1 (`Option<T>`)
                        # with a parameter of 4 where 4 indexes into the type table as `u32`.
                        "type": {
                            "generic": {
                                "type": 1,
                                "params": [ 3 ],
                            },
                        }
                    }
                ],
            }
        },
        { # ID 6
            # Demonstrates usage of a type that contains a Option<bool> type.
            "composite": {
                "path": [ 6 ],
                "fields": [
                    {
                        "type": {
                            # Demonstrates usage of the generic type at index 1 (`Option<T>`)
                            # with a parameter of 4 where 4 indexes into the type table as `bool`.
                            "generic": {
                                "type": 1,
                                "params": [ 4 ],
                            },
                        }
                    }
                ],
            }
        },
        { # ID 7
            # Demonstrates usage of a non-generic type.
            "composite": {
                "path": [ 7 ],
                "field": [
                    {
                        "type": {
                            # 3 indexes into the type table and refers to a non
                            # generic (aka concrete) type, in this case `u32`.
                            "concrete": 3
                        },
                    },
                    {
                        "type": {
                            # 4 indexes into the type table and refers to a non
                            # generic (aka concrete) type, in this case `bool`.
                            "concrete": 4
                        }
                    },
                ]
            }
        }
    }
}
```

ascjones:
Closes #6.

Previously generic types were represented in metadata with a type definition for each concrete instance.

For example `Option<T>`, there would be a copy of the definition of the whole type for `Option<bool>`, `Option<u32>`, and so on for all concrete instances.

When considering types with multiple type parameters, and where type parameters are used to parameterize generic fields e.g. `struct Foo<T> { a: Option<T> }`, it's possible to end up with many copies of the type definition but with different concrete parameters.

Not only will this result in bloated and redundant metadata, but we are losing information that is present at the syntactical level i.e. that all concrete instances of a type are constructed from the generic type itself. This relationship can be inferred but should be explicit in the data model. E.g. the use case of generating Rust types from the metadata, it would make identifying and generating generic types considerably simpler.

- [x] Make it compile
- [x] Fix tests
  - [x] Make json generic test cases pass
- [ ] Add parameterized field support to macros
  - [x] Structs and enums
  - [x] Tuples
  - [x] Arrays
  - [x] Refs
  - [ ] Refactoring, maybe extract the field building logic
- [ ] Refactor mercilessly
  - [ ] Find a better name for `RegistryType/InternedType`
- [ ] Consider how/whether to include associated type support in this PR
- [ ] Documentation
  - [ ] MetaType
  - [ ] Registry
  - [ ] InternedType
- [ ] PR Descriiption
  - [ ] Type Ids
 -  [ ] MetaType changes
  - [ ] Interned/Normalized Type


ascjones:
Shelving this, for large metadata files caused by monomorphization of generic types we can explore general compression tools.

PR #17, opened 2020-06-29
Remove string table, inline strings

ascjones:
Removes the string table following a suggestion from @seanyoung.

Improves human readability, but does increase the resulting raw serialized file size. However if file sizes become a problem we could reduce it with a generic compression tool.

Rel: https://github.com/paritytech/ink/issues/299

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,8 +24,8 @@
 //! # Registry
 //!
 //! At the heart of its functionality is the [`Registry`](`crate::Registry`) that acts as cache for
-//! known strings and types in order to efficiently deduplicate them and thus
-//! compactify the overall serialization.
+//! known types in order to efficiently deduplicate them and thus compactify the overall
+//! serialization.
 //!
 //! # Type Information
 //!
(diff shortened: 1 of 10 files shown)


seanyoung:
Very nice indeed, thanks for doing that

src/lib.rs
```diff
@@ -38,20 +38,20 @@
...
-//! The compact form is retrieved by the [`IntoCompact`](`crate::IntoCompact`) trait and internally used
-//! by the [`Registry`](`crate::Registry`) in order to convert the uncompact strings and types into
-//! their compact form.
+//! There is an uncompact form, called [`MetaForm`](`crate::form::MetaForm`) that acts as a bridge
+//! from compile-time type information at runtime in order to easily retrieve all information needed
+//! to uniquely identify types.
+//!
+//! The compact form is retrieved by the [`IntoCompact`](`crate::IntoCompact`) trait and internally
+//! used by the [`Registry`](`crate::Registry`) in order to convert the uncompact strings and types
+//! into their compact form.
```
Robbepop, line 47:
```suggestion
//! The compact form is retrieved by the [`IntoCompact`](`crate::IntoCompact`) trait and internally
//! used by the [`Registry`](`crate::Registry`) in order to convert the uncompact types
//! into their compact form.
```

src/lib.rs
```diff
@@ -38,20 +38,20 @@
...
 //!
 //! # Symbols and Namespaces
 //!
-//! Since symbol names are often shared across type boundaries the [`Registry`](`crate::Registry`)
-//! also deduplicates them. To differentiate two types sharing the same name
-//! namespaces are used. Commonly the namespace is equal to the one where the
-//! type has been defined in. For Rust prelude types such as [`Option`](`std::option::Option`) and
-//! [`Result`](`std::result::Result`)  the root namespace (empty namespace) is used.
+//! To differentiate two types sharing the same name namespaces are used.
+//! Commonly the namespace is equal to the one where the ype has been defined in. For Rust prelude
```
Robbepop, line 52:
```suggestion
//! Commonly the namespace is equal to the one where the type has been defined in. For Rust prelude
```

PR #19, opened 2020-08-07
Implement SCALE encoding and serde deserialization

### LINKED ISSUE #4: Add Registry Deserialization
See: https://github.com/type-metadata/type-metadata/pull/38, though this will require some work to adjust once #3 is merged.

ascjones:
Closes #4 (rel https://github.com/type-metadata/type-metadata/pull/38).

Introduces `OwnedForm` and `RegistryReadOnly` to allow SCALE decoding and JSON deserialization.

- [ ] Consider adding features for `scale` and `serde`, maybe decoding specific for `no_std`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ std = [
 	"scale/std"
 ]
 derive = [
-    "scale-info-derive"
+	"scale-info-derive"
 ]
 
 [workspace]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ scale = { package = "parity-scale-codec", version = "1.3.4", default-features = 
 [features]
 default = ["std"]
 std = [
-    "serde/std",
+	"serde/std",
 	"scale/std"
 ]
 derive = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,11 +17,13 @@ include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 scale-info-derive = { version = "0.2.0", path = "derive", default-features = false, optional = true }
 serde = { version = "1", default-features = false, features = ["derive", "alloc"] }
 derive_more = { version = "0.99.1", default-features = false, features = ["from"] }
+scale = { package = "parity-scale-codec", version = "1.3.4", default-features = false, features = ["derive"] }
 
 [features]
 default = ["std"]
 std = [
     "serde/std",
+	"scale/std"
 ]
 derive = [
     "scale-info-derive"
--- a/src/form.rs
+++ b/src/form.rs
@@ -72,16 +72,5 @@ pub enum CompactForm {}
 
 impl Form for CompactForm {
 	type TypeId = UntrackedSymbol<TypeId>;
-	type String = &'static str;
-}
-
-/// Owned form for decoding
-#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Serialize, Debug)]
-pub enum OwnedForm {}
-
-impl Form for OwnedForm {
-	/// For correctness should be [`NonZeroU32`](`core::num::NonZeroU32`), but the current
-	/// (`1.3.x`) release of `parity-scale-codec` does not include the `NoneZero*` Decode impls.
-	type TypeId = u32;
 	type String = String;
 }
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -17,11 +17,13 @@ include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 scale-info-derive = { version = "0.2.0", path = "derive", default-features = false, optional = true }
 serde = { version = "1", default-features = false, features = ["derive", "alloc"] }
 derive_more = { version = "0.99.1", default-features = false, features = ["from"] }
+scale = { package = "parity-scale-codec", version = "1.3.4", default-features = false, features = ["derive"] }
 
 [features]
 default = ["std"]
 std = [
     "serde/std",
+	"scale/std"
```
Robbepop, line 26: strange indentation

src/form.rs
```diff
@@ -42,6 +42,8 @@ use serde::Serialize;
 pub trait Form {
 	/// The type identifier type.
 	type TypeId: PartialEq + Eq + PartialOrd + Ord + Clone + core::fmt::Debug;
+	/// The string type.
+	type String: Serialize + PartialEq + Eq + PartialOrd + Ord + Clone + core::fmt::Debug;
```
Robbepop, line 46: While I like the overall idea I think it is better in this case to opt-into KISS (keep it simple stupid) and instead convert all `&'static str` simply into `String`. The `&'static str` was used in the beginning when we did not need deserialization in ink!. Now that we want to support Substrate we need it. However, if we do not generate metadata in any performance critical code section we don't need to opt-in for maximum performance and can instead opt-into simple code.

src/form.rs
```diff
@@ -62,9 +65,23 @@ impl Form for MetaForm {
...
+}
+
+/// Owned form for decoding
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Serialize, Debug)]
+pub enum OwnedForm {}
+
+impl Form for OwnedForm {
+	/// For correctness should be [`NonZeroU32`](`core::num::NonZeroU32`), but the current
+	/// (`1.3.x`) release of `parity-scale-codec` does not include the `NoneZero*` Decode impls.
+	type TypeId = u32;
```
Robbepop, line 85: We should not use raw types for indices but instead something type safe using a new-type wrapper. However, this doesn't matter too much if we remove `OwnedForm` again due to my proposal above.

src/registry.rs
```diff
@@ -148,3 +166,17 @@ impl Registry {
...
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Decode)]
+pub struct RegistryReadOnly {
+	types: Vec<Type<OwnedForm>>,
+}
+
+impl RegistryReadOnly {
+	/// Returns the type definition for the given identifier, `None` if no type found for that ID.
+	pub fn resolve(&self, id: NonZeroU32) -> Option<&Type<OwnedForm>> {
+		self.types.get((id.get() - 1) as usize);
+		None
```
Robbepop, line 180: AS far as I see this always returns `None`. Is that correct?

src/ty/mod.rs
```diff
@@ -95,25 +99,29 @@ impl Type {
 			path,
 			type_params: type_params.into_iter().collect(),
 			type_def: type_def.into(),
+			// marker: MetaForm,
```
Robbepop, line 102: ?

test_suite/derive_tests_no_std/src/main.rs
```diff
@@ -14,48 +14,8 @@
...
-	fn ext_free(ptr: *mut u8);
-}
-unsafe impl GlobalAlloc for Allocator {
-	unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
-		ext_malloc(layout.size()) as *mut u8
-	}
-	unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
-		ext_free(ptr as *mut u8)
-	}
-}
```
Robbepop, line 58: bless god this is gone :sweat_smile:

src/registry.rs
ascjones: I see you spotted my deliberate mistake :sweat_smile:

src/form.rs
ascjones: I've removed the `OwnedForm` now, but kept `MetaForm` as `& 'static str`, and changed `CompactForm` to `String`. Let me know what you think.

PR #20, opened 2020-09-08
Derive TypeInfo for fields with associated types without bounds

ascjones:
*WIP*, copied and amended `trait_bounds` from `parity_scale_codec`. Works but needs refactoring and documenting.

e.g.

```
trait Types {
	type A; // doesn't require TypeInfo trait bound
}
#[derive(TypeInfo)]
struct Assoc<T: Types> {
	a: T::A 
}

#[derive(TypeInfo)]
enum ConcreteTypes {}
impl Types for ConcreteTypes {
	type A = bool;
}
```

This generates a TypeInfo impl with bounds on the associated types i.e.

```
impl<T> for Assoc<T> where T: Types, T::A: TypeInfo
```


dvdplm:
Looks like this is in a pretty good state? What is still outstanding here, docs and [that todo](https://github.com/paritytech/scale-info/blob/62a28e35113b945d9d327c2e8f1e1b25f8db2eb7/derive/src/trait_bounds.rs#L151)?

ascjones:
Ooh thanks for pinging this. Yeah just the things you specified there and syncing with master

derive/src/trait_bounds.rs
```diff
@@ -0,0 +1,197 @@
...
+			visit::visit_type_path(self, i);
+		} else {
+			self.result.push(i.clone());
+		}
+	}
+}
+
+/// Collects all type paths that do not start or contain the given ident in the given type.
+///
+/// Returns `T`, `N`, `A` for `Vec<(Recursive<T, N>, A)>` with `Recursive` as ident.
```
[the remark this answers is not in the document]
ascjones: Collects all the child idents `["Recursive", "T", "N", "A"]`, and filters out the specified in this case "Recusive". I agree it's not very clear, will improve this comment

test_suite/tests/derive.rs
```diff
@@ -179,3 +179,120 @@ fn fields_with_type_alias() {
...
+    #[allow(unused)]
+    #[derive(TypeInfo)]
+    struct Meee {
+        me: Box<Meee>,
+    }
+    fn assert_type_info<T: TypeInfo + 'static>() {};
+    assert_type_info::<Meee>();
+}
+
+#[rustversion::nightly]
```
dvdplm, line 291: Error messages are quite different on stable currently and as we all (?) use nightly for development I went for this.

test_suite/tests/ui/fail_non_static_lifetime.rs
```diff
@@ -0,0 +1,12 @@
...
+
+#[derive(TypeInfo)]
+struct Me<'a> {
+    me: &'a Me<'a>,
+}
+
+fn assert_type_info<T: TypeInfo + 'static>() {}
+
+fn main() {
+    assert_type_info::<Me>();
```
ascjones, line 11: What we really want is that `assert_type_info::<Me<'static>()` works, with an impl looking like:

```
impl TypeInfo for Me<'static> {
    type Identity = Self;
    fn type_info() -> scale_info::Type {
        // todo: .field_of<&'static Me<'static>>,
        unimplemented!()
    }
}
```

However I realise this is out of the scope of this PR. I did something like this in my closed generics PR: https://github.com/paritytech/scale-info/pull/12/files#diff-60da9bde79deaa50c0ce6f42d1f47205ba31f448dc2713cc746339343581b699R113

derive/src/trait_bounds.rs
```diff
@@ -0,0 +1,160 @@
...
+            }
+        }
+
+        Data::Enum(ref data) => {
+            data.variants
+                .iter()
+                .filter(|variant| variant_filter(variant))
+                .flat_map(|variant| {
+                    match &variant.fields {
+                        Fields::Named(FieldsNamed { named: fields, .. })
```
ascjones, line 134: Can this code block be deduplicated with the equivalent in lines 113 to 124, or not worth it you think?

derive/src/trait_bounds.rs
```diff
@@ -0,0 +1,160 @@
...
+    collect_types(
+        &data,
+        type_filter,
+        // All types that show up in enum variants are filtered with the same
+        // type filter as regular types.
+        |_| true,
+    )
+}
+
+fn collect_types(
```
ascjones, line 103: It appears `collect_types` is only used from `collect_types_to_bind`, so they could be merged to a single function

derive/src/trait_bounds.rs
```diff
@@ -0,0 +1,160 @@
...
+
+/// Checks if the given type contains one of the given idents.
+fn type_contains_idents(ty: &Type, idents: &[Ident]) -> bool {
+    let mut visitor = ContainIdents {
+        result: false,
+        idents,
+    };
+    visitor.visit_type(ty);
+    visitor.result
+}
```
ascjones, line 48: I would prefer the `add` function below to be the top of the file as the entry point, helpers below. Also this visitor might be better to be inline inside the only function where it is used, or at least just below it for easy reference

derive/src/trait_bounds.rs
```diff
@@ -0,0 +1,160 @@
...
+
+/// Adds a `TypeInfo + 'static` bound to all relevant generic types, correctly
```
ascjones, line 50: I would add that it adds bounds on associated types e.g. `T::A: TypeInfo`

derive/src/trait_bounds.rs
```diff
@@ -0,0 +1,160 @@
...
+    if ty_params.is_empty() {
+        return Ok(())
+    }
+
+    let types = collect_types_to_bind(input_ident, data, &ty_params)?;
+
+    if !types.is_empty() {
+        let where_clause = generics.make_where_clause();
+
+        types.into_iter().for_each(|ty| {
```
ascjones, line 69: An alternative API would be to return these types and have the caller add them to the where clause, rather than passing `&mut Generics`. Not saying it's better just thinking out loud.

derive/src/trait_bounds.rs
dvdplm: definitely

derive/src/trait_bounds.rs
dvdplm: Yeah it's a toss-up. Passing around a `&mut` is a bit iffy, but keeps the caller short(er). Moving this (and another snippet) to the caller would make the `trait_bounds` be less coupled to `TypeInfo` stuff which might be valuable if we ever need similar kinds of type filtering for other purposes.
I'd be inclined to keep it as-is for now, mostly because I'm lazy; if you disagree I'll move it.

ascjones: Fine by me

test_suite/tests/ui/fail_non_static_lifetime.rs
dvdplm: Opened https://github.com/paritytech/scale-info/issues/38

test_suite/tests/derive.rs
```diff
@@ -179,3 +179,120 @@ fn fields_with_type_alias() {
...
+        ears: [Ear; 3],
+        paws: PawType<Paw>,
+    }
+
+    fn assert_type_info<T: TypeInfo + 'static>() {};
+    assert_type_info::<Cat<bool, u8, u16>>();
+}
+
+#[test]
+fn adds_correct_bounds_to_self_referential_types() {
```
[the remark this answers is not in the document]
dvdplm: Moving to trybuild test.

test_suite/tests/derive.rs
```diff
@@ -179,3 +179,120 @@ fn fields_with_type_alias() {
...
+fn adds_proper_trait_bounds() {
+    #[allow(unused)]
+    #[derive(TypeInfo)]
+    struct Cat {
+        tail: bool,
+        ears: u8,
+    }
+    fn assert_type_info<T: TypeInfo + 'static>() {};
+
+    assert_type_info::<Cat>();
```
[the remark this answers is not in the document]
dvdplm: Good idea, doing that.

ascjones (commented):
LGTM

dvdplm (approved):
Approval by proxy

PR #21, opened 2020-09-09
Derive TypeInfo impls for meta type definitions

ascjones:
#19 enables scale encoding of the `scale-info` type definitions. In order to decode these from Rust, a client can add a dependency to the `scale-info` crate. 

However, other languages will need to define their own types which will need to be handwritten. e.g. https://github.com/polkadot-js/api/blob/a67ffce84736e12216721d7191aadba47d9944f8/packages/types/src/interfaces/contractsAbi/definitions.ts#L76.

This PR adds a `dogfood` feature (may be renamed) which will generate `TypeInfo` implementations for the `scale_info::Type` types themselves. This will enable generating equivalent types in other languages which can be used for decoding.

Depends on #20.


ascjones:
Superseded by #72

PR #22, opened 2020-09-09
Add public getters for fields in meta type hierarchy

ascjones:
Following on from #19 which enables decoding of `scale-info` meta types, this PR makes the fields of those types accessible via public getters.

The motivation for this is to allow 3rd party libraries to make use of instances of the meta types for e.g code generation of types in another language.

The immediate use case I am planning for is to build a library which is able to deserialize `ink!` contract metadata and use that to construct contract calls.

**Additionally** we rename `Form::TypeId` to `Form::Type` to be more accurate.

### DIFF
--- a/src/ty/mod.rs
+++ b/src/ty/mod.rs
@@ -33,8 +33,8 @@ pub use self::{composite::*, fields::*, path::*, variant::*};
 /// A [`Type`] definition with optional metadata.
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, From, Debug, Serialize, Deserialize, Encode, Decode)]
 #[serde(bound(
-	serialize = "T::TypeId: Serialize, T::String: Serialize",
-	deserialize = "T::TypeId: DeserializeOwned, T::String: DeserializeOwned"
+	serialize = "T::Type: Serialize, T::String: Serialize",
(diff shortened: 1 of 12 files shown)


Robbepop (commented):
Is this not possible with some idiomatic getters instead?

ascjones:
I thought it was idiomatic for these types which are data, e.g. in `syn`?

src/ty/mod.rs
```diff
@@ -103,6 +103,26 @@ impl Type {
...
+
+	/// Returns the generic type parameters of the type
+	pub fn type_params(&self) -> &[T::Type] {
+		&self.type_params
+	}
+
+	/// Returns the definition of the type
+	pub fn type_def(&self) -> &TypeDef<T> {
+		&self.type_def
+	}
```
Robbepop, line 123: Why getters for `Type` whereas you have `pub` fields everywhere else?

Robbepop: Not that I dislike this. :sweat_smile:

Robbepop (approved):
LGTM

src/ty/mod.rs
ascjones: Actually `pub` fields are gone, replaced by getters

PR #24, opened 2020-10-11
Add extern crate proc_macro; release v0.4.1

ascjones:
/cc @Robbepop

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scale-info"
-version = "0.4.0"
+version = "0.4.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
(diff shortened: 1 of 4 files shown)


Robbepop (approved):

Robbepop:
Thanks!

Cargo.toml
```diff
@@ -14,7 +14,7 @@ categories = ["no-std", "encoding"]
 include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 
 [dependencies]
-scale-info-derive = { version = "0.2.0", path = "derive", default-features = false, optional = true }
+scale-info-derive = { version = "0.4.0", path = "derive", default-features = false, optional = true }
```
Robbepop, line 17: I think a version increment to 0.2.1 would be better.

Robbepop: So if an ink! contract depend on version `0.2` it would not need to update.

PR #26, opened 2020-11-12
Unify transparent wrapper types e.g. references

ascjones:
## Background

A bug was discovered by @Robbepop with the following recursive type using `Box`:

```
#[derive(TypeInfo)]
pub enum Tree {
	Leaf { value: i32 },
	Node { right: Box<Tree>, left: Box<Tree> },
}
```

This should result in a self-referential type definition in the type registry, instead it was creating two equivalent definitions: `Tree` and `Box<Tree>`.

The cause is that `any::TypeId::of<T>()` is used to uniquely identify types, and `Tree` and `Box<Tree>` are different types and therefore have different `TypeId`s. This regression was likely introduced in https://github.com/paritytech/scale-info/pull/3, which removed the original separation between a type id and its definition.

## "Transparent" wrapper types in SCALE

`Box` is an example of what in `parity-scale-codec` is known as a "wrapper" type. Such types are marked by the [`WrapperTypeEncode`](https://docs.rs/parity-scale-codec/1.3.5/parity_scale_codec/trait.WrapperTypeEncode.html) trait:

> A marker trait for types that wrap other encodable type. Such types should not carry any additional information that would require to be encoded, because the encoding is assumed to be the same as the wrapped type.

For our purposes that means that the SCALE encoded representation of `T` is the same as that of `Box<T>`, therefore we consider it a "transparent" wrapper type. Such a type would be erased in the type registry since it has the same definition.

Other types which implement `WrapperTypeEncode` are 

image

### Faux specialization

**EDIT: I abandoned this approach because it doesn't work with generics e.g. `GenericStruct<Box<GenericStruct>>`**

Therefore what we need is to be able to say that `meta_type(T) == metatype(Box<T>) == metadtype(&T) ...)`. To do this, for any type which implements `WrapperTypeEncode` we need to identify by `any::TypeId::of<T>()` (the id of the wrapped type). This way the same type definition is used for the wrapped type and it is erased during type registration.

The initial approach in this PR is adapted from the [autoref based specialization](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md). Here is a [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4a1f4b737f6894d0480a9b2d76fec472) demonstrating the concept.

### Adding an associated type to the `TypeInfo` trait

This is a simple alternative, which adds the associated type `MetaType` to `TypeInfo`. This will be set to `Self` by the derive macro and for most primitive type definitions. For the transparent wrapper types described above it will be set to the underlying wrapped type e.g. the `T` in `Box<T>`. 

The associated type will then be used when constructing the metatype to get the unique id with `core::any::TypeId::of<T::MetaType>()`. So for any type `T`, the equivalent `Box<T>` will have the same type id, and there will appear only a single type definition in the type registry.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -71,6 +71,7 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
 
 	let type_info_impl = quote! {
 		impl #impl_generics _scale_info::TypeInfo for #ident #ty_generics #where_clause {
+			type MetaType = Self;
 			fn type_info() -> _scale_info::Type {
 				_scale_info::Type::builder()
 					.path(_scale_info::Path::new(stringify!(#ident), module_path!()))
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -93,11 +93,11 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
 			let (ty, ident) = (&f.ty, &f.ident);
 			if let Some(i) = ident {
 				quote! {
-					.field_of::<#ty>(stringify!(#i))
+					.field(stringify!(#i), _scale_info::meta_type!(#ty))
 				}
 			} else {
 				quote! {
-					.field_of::<#ty>()
+					.field(_scale_info::meta_type!(#ty))
 				}
 			}
 		})
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -93,11 +93,11 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
 			let (ty, ident) = (&f.ty, &f.ident);
 			if let Some(i) = ident {
 				quote! {
-					.field(stringify!(#i), _scale_info::meta_type!(#ty))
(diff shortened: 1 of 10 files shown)


derive/src/lib.rs
```diff
@@ -71,6 +71,7 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
 
 	let type_info_impl = quote! {
 		impl #impl_generics _scale_info::TypeInfo for #ident #ty_generics #where_clause {
+			type MetaType = Self;
```
Robbepop, line 74: Could we maybe rename this to something less generic? E.g. `UnderlyingType`?

Robbepop: Or maybe `WrappedType` but I like `UnderlyingType` more tbh.

Robbepop (commented):
LGTM. I like the simplicity of this solution.
The only thing I'd like is to rename the generic `MetaType` to something more concrete such as `UnderlyingType`, `IdentityType` or just `Identity` something that really describes its semantics.

Robbepop (approved):

PR #27, opened 2020-11-19
Enumerate RegistryReadOnly types, Display Path

ascjones:
This is required for https://github.com/paritytech/cargo-contract/pull/79, in order to index the types by the `Path` so we can identify common environment types for custom encoding/decoding.

**Bonus:** Adds `Display` impl for `Path`

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -194,4 +194,35 @@ impl RegistryReadOnly {
 	pub fn resolve(&self, id: NonZeroU32) -> Option<&Type<CompactForm>> {
 		self.types.get((id.get() - 1) as usize)
 	}
+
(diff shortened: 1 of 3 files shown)


src/registry.rs
```diff
@@ -194,4 +194,35 @@ impl RegistryReadOnly {
...
+		})
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn readonly_enumerate() {
```
Robbepop, line 212: maybe another test with some recursive data structures would be nice

Robbepop (approved):
since the other PR for recursive data types has not yet been merge I will simply approve

PR #28, opened 2020-11-20
Minor cleanup

dvdplm:
Did some code reading and did some light editing as I went. And a little code change as well for good measure. :)


ascjones (commented):
Thanks!

ascjones (approved):

PR #30, opened 2020-11-24
Add informational field type name

ascjones:
Given a complex type which incorporates a field with a type alias, we currently lose the information about the use of a type alias. e.g.

```
type Balance = u128

#[derive(TypeInfo)]
struct S { b: Balance }
```

This currently results in the field type being resolved to `u128` in the type registry, and we have no way to distinguish between a plain `u128` and an aliased `u128`, which we may want to treat differently, e.g. some custom decoding or display logic for that type.

This PR adds the optional `type_name` property to `Field`, and updates the derive macro to extract the field type names e,g, in the above example the underlying type is `u128` but the type name is `Balance`.

Note that this will add a `type_name` even if it is non-aliased, e.g. `a: u32` will still have the type display name `"u32"` as part of the field definition. Even though this is redundant, and it is possible to remove non aliased display names (I implemented then removed it), I have left them in for the sake of simplicity and flexibility, leaving it up to consumers to do the lookup to determine whether an alias is being used or not.

This is required for https://github.com/paritytech/cargo-contract/pull/79 in order for custom encoding/decoding for aliased `Balance` types.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -19,7 +19,7 @@ extern crate proc_macro;
 
 mod impl_wrapper;
 
-use alloc::vec::Vec;
+use alloc::{string::ToString, vec::Vec};
 use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
@@ -91,15 +91,39 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
 		.iter()
 		.map(|f| {
 			let (ty, ident) = (&f.ty, &f.ident);
-			if let Some(i) = ident {
-				quote! {
-					.field_of::<#ty>(stringify!(#i))
-				}
-			} else {
-				quote! {
-					.field_of::<#ty>()
-				}
-			}
+
+			let display_name =
+				if let syn::Type::Path(type_path) = ty {
+					if type_path.qself.is_some() || type_path.path.segments.is_empty() {
+						quote! {}
+					} else {
+						let segs = type_path
+							.path
+							.segments
+							.iter()
+							.map(|seg| seg.ident.to_string())
+							.collect::<Vec<_>>();
+						quote! {
+							.with_type_display_name(
+								vec![#(#segs),*].into_iter().map(AsRef::as_ref)
+							)
+						}
+					}
+				} else {
+					quote! {}
+				};
+
+			let field =
+				if let Some(i) = ident {
+					quote! {
+						_scale_info::Field::named_of::<#ty>(stringify!(#i))#display_name
+					}
+				} else {
+					quote! {
+						_scale_info::Field::unnamed_of::<#ty>()#display_name
+					}
+				};
+			quote! { .field(#field) }
 		})
 		.collect()
 }
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -19,7 +19,13 @@ extern crate proc_macro;
 
 mod impl_wrapper;
 
-use alloc::vec::Vec;
+use alloc::{
+    string::{
+        String,
+        ToString,
+    },
+    vec::Vec,
+};
 use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
@@ -104,18 +110,36 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
         .iter()
         .map(|f| {
             let (ty, ident) = (&f.ty, &f.ident);
+            let type_name = clean_type_string(&quote!(#ty).to_string());
 
             if let Some(i) = ident {
                 quote! {
-                    .field_of::<#ty>(stringify!(#i), stringify!(#ty))
+                    .field_of::<#ty>(stringify!(#i), #type_name)
                 }
             } else {
                 quote! {
-                    .field_of::<#ty>(stringify!(#ty))
+                    .field_of::<#ty>(#type_name)
                 }
             }
         })
         .collect()
+}
+
+fn clean_type_string(input: &str) -> String {
+    input
+        .replace(" ::", "::")
+        .replace(":: ", "::")
+        .replace(" ,", ",")
+        .replace(" ;", ";")
+        .replace(" [", "[")
+        .replace("[ ", "[")
+        .replace(" ]", "]")
+        .replace(" (", "(")
+        .replace("( ", "(")
+        .replace(" )", ")")
+        .replace(" <", "<")
+        .replace("< ", "<")
+        .replace(" >", ">")
 }
 
 fn generate_composite_type(data_struct: &DataStruct) -> TokenStream2 {
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -105,7 +105,7 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
 						.collect::<Vec<_>>();
 					quote! {
 						.with_type_display_name(
-							vec![#(#segs),*].into_iter().map(AsRef::as_ref)
+							__core::vec![#(#segs),*].into_iter().map(AsRef::as_ref)
 						)
 					}
 				}
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -19,10 +19,7 @@ extern crate proc_macro;
 
 mod impl_wrapper;
 
-use alloc::{
-    string::ToString,
-    vec::Vec,
-};
+use alloc::vec::Vec;
 use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
@@ -108,36 +105,15 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
         .map(|f| {
             let (ty, ident) = (&f.ty, &f.ident);
 
-            let display_name = if let syn::Type::Path(type_path) = ty {
-                if type_path.qself.is_some() || type_path.path.segments.is_empty() {
-                    quote! {}
-                } else {
-                    let segs = type_path
-                        .path
-                        .segments
-                        .iter()
-                        .map(|seg| seg.ident.to_string())
-                        .collect::<Vec<_>>();
-                    quote! {
-                        .with_type_display_name(
-                            __core::vec![#(#segs),*].into_iter().map(AsRef::as_ref)
-                        )
-                    }
-                }
-            } else {
-                quote! {}
-            };
-
-            let field = if let Some(i) = ident {
+            if let Some(i) = ident {
                 quote! {
-                    _scale_info::Field::named_of::<#ty>(stringify!(#i))#display_name
+                    .field_of::<#ty>(stringify!(#i), stringify!(#ty))
                 }
             } else {
                 quote! {
-                    _scale_info::Field::unnamed_of::<#ty>()#display_name
+                    .field_of::<#ty>(stringify!(#ty))
                 }
-            };
-            quote! { .field(#field) }
+            }
         })
         .collect()
 }
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -93,37 +93,35 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
 		.map(|f| {
 			let (ty, ident) = (&f.ty, &f.ident);
 
-			let display_name =
-				if let syn::Type::Path(type_path) = ty {
-					if type_path.qself.is_some() || type_path.path.segments.is_empty() {
-						quote! {}
-					} else {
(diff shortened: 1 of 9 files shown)


src/build.rs
```diff
@@ -271,6 +264,50 @@ impl FieldsBuilder<UnnamedFields> {
...
+	pub fn new(ty: MetaType) -> Self {
+		Self {
+			name: None,
+			ty,
+			ty_display_name: None,
+			marker: PhantomData,
+		}
+	}
+
+	/// Specify the display name of the type of the field. This can either be the type name itself
```
dvdplm, line 286: Why not name it "type alias"? "Display" seems like something used in a UI, but I don't think that's the only use case here. Can these be anything but type aliases?

src/build.rs
```diff
@@ -271,6 +264,50 @@ impl FieldsBuilder<UnnamedFields> {
...
+	pub fn with_type_display_name<S>(mut self, path_segments: S) -> Self
+	where
+		S: IntoIterator<Item = &'static str>,
+	{
+		let path = Path::from_segments(path_segments).expect("display name is invalid");
+		self.ty_display_name = Some(path);
+		self
+	}
+
+	/// Complete building the field.
```
dvdplm, line 297:
```suggestion
	/// Complete building the field. Consumes `self`.
```

test_suite/tests/derive.rs
```diff
@@ -50,7 +50,11 @@ fn struct_derive() {
 	let struct_type = Type::builder()
 		.path(Path::new("S", "derive"))
 		.type_params(tuple_meta_type!(bool, u8))
-		.composite(Fields::named().field_of::<bool>("t").field_of::<u8>("u"));
+		.composite(
+			Fields::named()
+				.field(Field::named_of::<bool>("t").with_type_display_name(vec!["T"]))
+				.field(Field::named_of::<u8>("u").with_type_display_name(vec!["U"])),
```
dvdplm, line 56: It's a bit awkward to have to pass in a `Vec` to `with_type_display_name` but I guess it has to be a full path. Maybe you could allow `with_type_display_name("full::path::to::alias::T")` and split it up in `with_type_display_name()`? But that would likely mess up things elsewhere right?

test_suite/tests/derive.rs
```diff
@@ -121,8 +129,14 @@ fn enum_derive() {
...
-				.variant("A", Fields::unnamed().field_of::<bool>())
-				.variant("B", Fields::named().field_of::<bool>("b"))
+				.variant(
+					"A",
+					Fields::unnamed().field(Field::unnamed_of::<bool>().with_type_display_name(vec!["T"])),
+				)
+				.variant(
+					"B",
+					Fields::named().field(Field::named_of::<bool>("b").with_type_display_name(vec!["T"])),
+				)
```
dvdplm, line 139: This was a bit more ergonomic before; having to repeat the info about building a `named`/`unnamed` field is a bit unfortunate. I wonder if `Fields::field(Field::unnamed_of::<bool>().with_type_display_name(…)` could work? 

And what if `Field::unnamed_of` was a free standing function, then it would be just `Fields::field(unnamed_of::<bool> … …)`?

test_suite/tests/derive.rs
```diff
@@ -140,14 +154,34 @@ fn recursive_type_derive() {
...
+					.field(Field::named_of::<Box<Tree>>("left").with_type_display_name(vec!["Box"])),
 			),
 	);
 
 	assert_type!(Tree, ty);
 }
+
+#[test]
+fn fields_with_type_alias() {
+	type BoolAlias = bool;
```
dvdplm, line 174: Unrelated to this PR, but do you think we could have this work at some point?

```
#[derive(TypeInfo)]
type BoolAlias = bool;
```

dvdplm (commented):
Two main concerns (really: nitpicks) are:
1. the API is a bit less ergonomic to use after this
2. not a fan of the `display_name` name; could it be just `alias`?

Overall great stuff though.

test_suite/tests/derive.rs
ascjones: It's a good idea, it may well be worth experimenting with. This would provide extra information about what aliases are available and may increase the accuracy of the heuristic identifying alias usage. 

Ultimately it would also have the same limitation in that there is still no guarantee to identify when a given alias is being used, because the registry is using the underlying type id i.e. `any::TypeId::of<BoolAlias>() == any::TypeId::of<bool>()`.

derive/src/lib.rs
```diff
@@ -92,15 +92,37 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
...
+			let display_name = if let syn::Type::Path(type_path) = ty {
+				if type_path.qself.is_some() || type_path.path.segments.is_empty() {
+					quote! {}
+				} else {
+					let segs = type_path
+						.path
+						.segments
+						.iter()
+						.map(|seg| seg.ident.to_string())
+						.collect::<Vec<_>>();
```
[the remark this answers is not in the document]
ascjones: Not that I'm aware of

src/ty/fields.rs
```diff
@@ -31,13 +32,20 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
...
+#[serde(rename_all = "camelCase")]
 pub struct Field<T: Form = MetaForm> {
 	/// The name of the field. None for unnamed fields.
 	#[serde(skip_serializing_if = "Option::is_none", default)]
 	name: Option<T::String>,
 	/// The type of the field.
 	#[serde(rename = "type")]
 	ty: T::Type,
+	/// The compile-time known displayed representation of the type of the field. This will be the
+	/// actual name of the type or an alias of it.
```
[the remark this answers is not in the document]
ascjones: Comment replaced

src/ty/fields.rs
```diff
@@ -31,13 +32,20 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
...
+	///
+	/// Will be `None` if the type has a qualified type path e.g. `<T as Trait>::AssociatedItem`.
```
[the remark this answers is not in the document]
ascjones: Latest commits remove this limitation. I copied the original idea from `ink!` which only supports non generic arguments.

src/build.rs
ascjones: Because it applies to any field type name, not only an alias.

test_suite/tests/derive.rs
ascjones: This is resolved now since the display name is no longer optional. The reason I made it this way was to optimize for composing the code in the macro (this code is ususally only written by hand in the tests), and to use a builder rather than creating many constructor permutations.

test_suite/tests/derive.rs
ascjones: Outdated now

ascjones:
Updated so it now contains the raw string which is a result of `stringify(#ty)` where `ty` is a `syn::Type`. Let me know what you think @Robbepop.

test_suite/tests/json.rs
```diff
@@ -171,9 +171,9 @@ fn test_tuplestruct() {
         "def": {
             "composite": {
                 "fields": [
-                    { "type": 1, "displayName": ["i32"] },
-                    { "type": 2 },
-                    { "type": 4, "displayName": ["bool"] },
+                    { "type": 1, "typeName": "i32" },
+                    { "type": 2, "typeName": "[u8 ; 32]" },
```
dvdplm, line 175: This is nice. :)

dvdplm (commented):
I'll give it a proper review, but this real good. :)

test_suite/tests/derive.rs
Robbepop: Never thought about this use case. I am not sure it goes too far but making it optional would probably be okay. The question is how to integrate it properly into the current scheme if we actually want ot.

test_suite/tests/json.rs
Robbepop: Note sure we need `displayName` or `typeName` (why rename btw?) for types such as `[u8; 32]` because its primary use case is still to counterfeit type aliases and besides that I cannot think of any other serious use case. Also I am unsure about turning the namespace (array of strings) into a single string since that kinda dictates how 3rd party tools handle namespaces or simply removes namespaces which are kinda important to differentiate between types.

Robbepop (commented):
already commented

test_suite/tests/json.rs
ascjones: > displayName or typeName (why rename btw?)

To make it more general to imply we are not just dealing with paths/aliases now, but whatever type name is in that position.

> Note sure we need displayName or typeName (why rename btw?) for types such as [u8; 32]

Indeed it is redundant because you can just look it up on the underlying type. We can easily filter those out, but in this iteration I have decided to always take the type name in the source for simplicity's sake, at the cost of some redundancy. 

> besides that I cannot think of any other serious use case

I believe we might need this initially if we want to be able to convert to the existing substrate metadata format, though of course it may be possible without, needs some investigation.

>  Also I am unsure about turning the namespace (array of strings) into a single string since that kinda dictates how 3rd party tools handle namespaces or simply removes namespaces which are kinda important to differentiate between types.

3rd party tools should not be relying on any namespace in the type/display name because it will only be there if the user decides to fully qualify their type e.g. `struct S { field: foo::bar::Foo }`, if they just `use foo::bar::Foo` then the display/type name will be `Foo`. 

By not using the `Path` type anymore we are just giving an extra indication that this field does *not* provide any guarantees as to the content of that field other than it is the name of a valid Rust type in the context of the `.rs` file where it is defined.

derive/src/lib.rs
```diff
@@ -104,20 +110,38 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
...
+        .replace(" ;", ";")
+        .replace(" [", "[")
+        .replace("[ ", "[")
+        .replace(" ]", "]")
+        .replace(" (", "(")
+        .replace("( ", "(")
+        .replace(" )", ")")
+        .replace(" <", "<")
+        .replace("< ", "<")
+        .replace(" >", ">")
```
Robbepop, line 142: :laughing:

dvdplm: dq: why wouldn't `.replace(" ", "");` work here? Is space ever significant?

Robbepop: yes, e.g. something like `&mut Foo` would become `&mutFoo`

dvdplm: 🤦

ascjones: I'd love to claim credit for this beauty but I stole it from https://github.com/paritytech/substrate/blob/master/frame/support/procedural/tools/src/lib.rs#L96

Robbepop:
The only thing that bogs me with this approach that we no longer really support overlapping aliases by replacing the namespace with plain strings. E.g.:

```rust
mod a {
    pub type Foo = i32;
}
mod b {
    pub type Foo = i32;
}

fn do_stuff(foo: a::Foo) { ... }
```

ascjones:
You can still differentiate between the two though, instead of `["a", "Foo"]` and `["b", "Foo"]` you'd just have `"a::Foo"` and `"b::Foo"`.

Robbepop:
> You can still differentiate between the two though, instead of `["a", "Foo"]` and `["b", "Foo"]` you'd just have `"a::Foo"` and `"b::Foo"`.

But then how is `"a::Foo"` string in the metadata an improvement over the more generic `["a", "Foo"]` array especially in case of our targeted cross-language ecosystem where not every supported language has the same Rust-y syntax for namespaces?

ascjones:
For the case of a plain `TypePath` like that it's not an improvement. a structured `Path` is clearly better. This was a way to get access to the unstructured type name for the field for any possible type without having to define a bunch of types to map to all possible variants of `syn::Type`. It's crude but KISS, and it provides enough information to do what I want for my use case.

However if we want something a bit more sophisticated it will just require a bit more thought. 

As an aside, I think it's useful information anyway. We probably don't want to model type in source code being `&mut <<T as MyTrait>::Type as MyOtherTrait>::Type`, all you care about is that it eventually resolves to `u64`, good to know what the source definition was.

Robbepop:
> For the case of a plain `TypePath` like that it's not an improvement. a structured `Path` is clearly better. This was a way to get access to the unstructured type name for the field for any possible type without having to define a bunch of types to map to all possible variants of `syn::Type`. It's crude but KISS, and it provides enough information to do what I want for my use case.
> 
> However if we want something a bit more sophisticated it will just require a bit more thought.
> 
> As an aside, I think it's useful information anyway. We probably don't want to model type in source code being `&mut <<T as MyTrait>::Type as MyOtherTrait>::Type`, all you care about is that it eventually resolves to `u64`, good to know what the source definition was.

ah okay I see.

Robbepop (approved):

dvdplm (approved):

PR #33, opened 2020-12-04
Consolidate common prelude for std and no_std and usage

ascjones:
Ideas copied from [ink!'s prelude](https://github.com/paritytech/ink/blob/master/crates/prelude/src/lib.rs).

- Fixes issue which required `extern crate alloc` to be present in root of `no_std` consumer
- Rename `tm_std` -> prelude
- Explicitly export common modules
- Use cfg_if for std and no_std environments
- Explicit prelude imports instead of `*`
- Use `::scale_info` in derive instead of `_scale_info::`
- Use `::scale_info` prelude from derive macro

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ categories = ["no-std", "encoding"]
 include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 
 [dependencies]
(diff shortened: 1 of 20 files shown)


dvdplm (approved):
LGTM

Tried this on a branch, seems like a solid cleanup.

PR #34, opened 2020-12-07
Optional serde feature

ascjones:
Makes the `serde` dependency optional, as per the convention. 

Integrating into the substrate runtime, we don't want or need a dependency on serde since it will be using SCALE encoding to marshal the metadata from the runtime.


dvdplm (commented):
LGTM

It would be good to document the available features in the `README` though. Arguably out of scope, but while you're in there, perhaps?

dvdplm (approved):

PR #35, opened 2020-12-07
Parameterize CompactForm String for optional SCALE impl

ascjones:
By default, parity-scale-codec does not provide Encode/Decode impls for an owned String.
This is only provided under the "full" feature which is not used by the substrate runtime,
because it should not be used for consensus critical code. So in order for the CompactForm
to be integrated into the substrate runtime, or wherever the "full" feature cannot be used,
then we must parameterize the `String` type so that it can be both an `&'static str` on the
runtime side where it is encoded, and a `String` in client/consuming code where it is decoded.

Note that previously I had done something similar with `OwnedForm` discussed with @Robbepop [here](https://github.com/paritytech/scale-info/pull/19#discussion_r472238289). But this or something similar is now required in case we want to generate the metadata directly inside the substrate runtime where `String`s are not allowed.

### DIFF
--- a/src/form.rs
+++ b/src/form.rs
@@ -34,6 +34,7 @@ use crate::prelude::{
     any::TypeId,
     fmt::Debug,
     marker::PhantomData,
+    string::String,
 };
 
 use crate::{
@@ -53,8 +54,17 @@ pub trait Form {
     /// The type representing the type.
     type Type: PartialEq + Eq + PartialOrd + Ord + Clone + Debug;
     /// The string type.
-    type String: PartialEq + Eq + PartialOrd + Ord + Clone + Debug;
+    type String: FormString;
 }
+
+/// Trait for types which can be used to represent strings in type definitions.
+pub trait FormString:
+    AsRef<str> + PartialEq + Eq + PartialOrd + Ord + Clone + Debug
+{
+}
+
+impl FormString for &'static str {}
+impl FormString for String {}
 
 /// A meta meta-type.
 ///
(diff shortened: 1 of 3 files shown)


src/form.rs
```diff
@@ -76,9 +76,12 @@ impl Form for MetaForm {
 ///
 /// `type String` is owned in order to enable decoding
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Serialize, Debug)]
-pub enum CompactForm {}
+pub struct CompactForm<S = &'static str>(PhantomData<S>);
```
dvdplm, line 79: For my education, can you elaborate on the choice of default here? Why is `&'static str` the natural choice here?

dvdplm (approved):
I can't say I fully understand the code surrounding your change here; the docs are a bit "academic" and I have a hard time following them at times. 
The changes in this PR are straight forward and seems like a no-brainer per se.

src/form.rs
ascjones: When constructing type infos the strings are always `'static`. Only for deserializing/decoding by consumers do we need owned strings.

ascjones:
I'd like @Robbepop to approve this too before merging, because he disagreed with my previous approach https://github.com/paritytech/scale-info/pull/19#discussion_r472238289

ascjones:
Related: https://stackoverflow.com/questions/65348505/runtime-building-string-not-found-in-this-scope

ascjones:
Merged since other work already based on this

PR #37, opened 2020-12-08
Doc tweaks

dvdplm:
…and some random nitpicks.

Diff is pretty noisy. Mergin master into `aj-bounds` likely helps with that.


derive/src/trait_bounds.rs
```diff
@@ -108,9 +108,13 @@ impl<'a, 'ast> Visit<'ast> for FindTypePathsNotStartOrContainIdent<'a> {
     }
 }
 
-/// Collects all type paths that do not start or contain the given ident in the given type.
+/// Collects all type paths that do not start or contain the given ident in the
```
ascjones, line 111:
```suggestion
/// Collects all type paths that do not start with or contain the given ident in the
```

ascjones (approved):

PR #39, opened 2020-12-14
Ensure only static lifetimes appear in derived types

### LINKED ISSUE #38: TypeInfo for types with 'static lifetime
Make this work:

```rust
impl TypeInfo for Me<'static> {
    type Identity = Self;
    fn type_info() -> scale_info::Type {
        // todo: .field_of<&'static Me<'static>>,
        unimplemented!()
    }
}
```

See the closed generics PR [here](https://github.com/paritytech/scale-info/pull/12/files#diff-60da9bde79deaa50c0ce6f42d1f47205ba31f448dc2713cc746339343581b699R113).

dvdplm:
Follow-up PR to #20 that attempts to fix #38


derive/src/trait_bounds.rs
```diff
@@ -28,29 +28,34 @@ use syn::{
...
+    if ty_params_ids.is_empty() {
         return Ok(())
     }
 
-    let types = collect_types_to_bind(input_ident, data, &ty_params)?;
-
-    if !types.is_empty() {
-        let where_clause = generics.make_where_clause();
+    let types = collect_types_to_bind(input_ident, data, &ty_params_ids)?;
+    let generics_clone = generics.clone();
```
dvdplm, line 41: This is a bit nasty; what's a better way?

ascjones: Could do something like `let type_params = generics.type_params().cloned().collect::<Vec<_>>();` and then use `type_params.into_iter().for_each(...` in line 50, could also remove the clones in the closure

test_suite/tests/derive.rs
```diff
@@ -199,7 +200,11 @@ fn associated_types_derive_without_bounds() {
     let struct_type = Type::builder()
         .path(Path::new("Assoc", "derive"))
         .type_params(tuple_meta_type!(ConcreteTypes))
-        .composite(Fields::named().field_of::<bool>("a", "T::A"));
+        .composite(
+            Fields::named()
+                .field_of::<bool>("a", "T::A")
+                .field_of::<u64>("b", "& \'static u64"),
```
ascjones, line 206: We may want to clean that space up

ascjones (approved):
LGTM

derive/src/trait_bounds.rs
```diff
@@ -28,29 +28,34 @@ use syn::{
...
-        });
-    }
+    types.into_iter().for_each(|ty| {
+        where_clause
+            .predicates
+            .push(parse_quote!(#ty : ::scale_info::TypeInfo + 'static))
+    });
 
+    type_params.iter().for_each(|type_param| {
+        let ident = type_param.ident.clone();
```
ascjones, line 51: These clones may be redundant now if you use `into_iter`

PR #40, opened 2020-12-15
Use PortableRegistry for encoding and serializing

ascjones:
Removes custom encoding/serialization from the `Registry` and instead uses the `PortableRegistry` which already provides a `From<Registry>` impl.

### DIFF
--- a/test_suite/tests/json.rs
+++ b/test_suite/tests/json.rs
@@ -32,7 +32,7 @@ use scale_info::{
     meta_type,
     IntoPortable as _,
     Registry,
-    RegistryReadOnly,
+    PortableRegistry,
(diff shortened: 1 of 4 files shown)


test_suite/tests/json.rs
```diff
@@ -296,6 +297,7 @@ fn test_recursive_type_with_box() {
         ]
     });
 
+    let registry: PortableRegistry = registry.into();
```
dvdplm, line 300: I guess you could do `serde_json::to_value::<PortableRegistry>(registry.into())` but that's not better.

dvdplm (approved):
LGTM

Issue #8, opened 2020-04-02
Support Compact types


ascjones:
I just had a chat with @jacogr yesterday and we came to the conclusion that we also need support for `Compact` types. We could implement those into our format in 3 different ways and I do not know which one is the best.
- We could provide a flag that states whether some type is compact encoded, similar to how it has been done for `displayName` in ink!.
- We could provide another built-in type alongside the others like `Slice` and `Array` that is `Compact` and has a single generic parameter.
- We could make it a custom type that has special meaning and a prelude namespace.

_Originally posted by @Robbepop in https://github.com/paritytech/scale-info/pull/3_

Robbepop:
Let me quickly concretize what those 3 different approaches could look like:

## Examples

Following are 3 `scale-info` encoded JSON example showing a composite type:

```rust
pub struct Foo(Compact<i32>);
```

Using a built-in `scale-info` `compact` type:

```
{
    "strings": [
        "Foo", # ID 1
    ],
    "types": [
        { # ID 1
            "compact": 2
        },
        { # ID 2
            "primitive": "i32"
        },
        { # ID 3
            "composite": {
                "path": [ 1 ],
                "fields": [
                    { "type": 1 }
                ]
            }
        }
    ]
}
```

Using a built-in flag for compact types:

```
{
    "strings": [
        "Foo", # ID 1
    ],
    "types": [
        { # ID 1
            "primitive": "i32"
        },
        { # ID 2
            "composite": {
                "path": [ 1 ],
                "fields": [
                    { "type": 1, "compact": true }
                ]
            }
        }
    ]
}
```

Using a special custom type for compact types: (Taking the one from `scale` library)

```
{
    "strings": [
        "scale",   # ID 1
        "Compact", # ID 2
        "T",       # ID 3
        "Foo",     # ID 4
    ],
    "types": [
        { # ID 1
            "primitive": "i32"
        },
        { # ID 2
            "composite": {
                "path": [ 1, 2 ],
                "params": [ 3 ],
                "fields": [
                    {
                        "type": {
                            "parameter": 3
                        }
                    }
                ]
            }
        },
        { # ID 3
            "composite": {
                "path": [ 1 ],
                "fields": [
                    {
                        "type": {
                            "generic": {
                                "type": 2,
                                "params": [ 1 ],
                            }
                        }
                    }
                ]
            }
        }
    ]
}
```

Robbepop:
[...]

PR #41, opened 2020-12-17
Rename Compact to Portable

dvdplm:
Based on #35 (aj-compact-string).

This PR changes the terminology from `compact` into `frozen`, which is useful to avoid confusion with the terminology used in `parity-scale-codec`, where the `Compact` type has a prominent role (c.f. #8).

"Frozen" is a sub-optimal name. Other candidates include: `Universal` (too vague), `Static` (confusing), `Standalone` (too vauge), `Decontexualized` (is that even a word?), `Fixed` (seems to imply something about space usage), `Exported` (that's not the key aspect though is it?), `Serializable` (kind of confusing with `serde` being in the mix).

Also included: a bunch of edits to the documentation in the hope of improving readability a bit.


ascjones:
What about "Portable"?

dvdplm:
`Portable` works for me. Not ideal either (and it's not a movie) but it's better than `Frozen`. @Robbepop any comment on this before I unleash `grep`?

Robbepop:
> `Portable` works for me. Not ideal either (and it's not a movie) but it's better than `Frozen`. @Robbepop any comment on this before I unleash `grep`?

Cannot come up with a better idea for another name. The names never bothered me anyways.

dvdplm:
@ascjones ptal

src/lib.rs
```diff
@@ -15,49 +15,53 @@
...
+//! `scale-info` provides implementations for all commonly used Rust standard
+//! types and a derive macro for implementing of custom types.
 //!
-//! # Compaction Forms
+//! # Forms
 //!
-//! There is an uncompact form, called [`MetaForm`](`crate::form::MetaForm`) that acts as a bridge
-//! from compile-time type information at runtime in order to easily retrieve all information needed
-//! to uniquely identify types.
+//! There is an expanded form, called [`MetaForm`](`crate::form::MetaForm`) that
```
ascjones, line 41: I don't think `expanded` is the right word here. That said, I'm not sure what the right word would be here either. Could maybe just remove `There is an expanded form` and just use the rest of the sentence. Or something to describe that this is how types are described and identified statically.

ascjones (approved):
LGTM

src/lib.rs
dvdplm: The whole paragraph is a bit weird imo. Hard to rewrite because I'm not sure I understand it myself. Changing to:

```
//! # Forms
//!
//! To bridge between compile-time type information and runtime the
//! [`MetaForm`](`crate::form::MetaForm`) is used to easily retrieve all
//! information needed to uniquely identify types.
```

dvdplm: Feel free to suggest moar betterments, happy to make the changes.

PR #42, opened 2020-12-17
Add a `compact` flag to `Field`

### LINKED ISSUE #8: Support Compact types

dvdplm:
Closes #8
Builds on #41

Adds a `compact` boolean to `Field` to signal that the type should be encoded/decoded as a `parity_scale_codec::Compact` type.

~~This is WIP and does not yet change the derive macro to use the new flag.~~

The code here takes the first approach outlined in #8, using a flag. The reasoning is to try the simplest thing that works and see where it takes us. The preferred approach outlined by @Robbepop, leveraging the existing machinery to derive `Compact<T>` is elegant but given the scary expansion of `#[codec(compact)]` (see comment below) I'm not sure if it is actually workable (tbh I haven't tried yet).


src/build.rs
```diff
@@ -248,6 +248,15 @@ impl<T> FieldsBuilder<T> {
...
     }
+
+    /// Mark last field as compact, meaning that encoding/decoding should be in the [`scale_codec::Compact`] format.
+    pub fn compact(mut self) -> Self {
+        self.fields.iter_mut().last().and_then(|f| {
+            f.compact();
+            Some(f)
+        });
+        self
+    }
```
dvdplm, line 259: @ascjones This is a bit terrible but when it is [used](https://github.com/paritytech/scale-info/pull/42/files#diff-2f1d807141fcd1651cee5c5469b400f87ad94fa2fef09135bb1d100355ee4fc3R225) it seems to work out fine. The alternative is to add a new parameter to `field_of()` but it's a bit opaque for readers: `Fields::named().field_of::<[u8; 32]>("thing", "[u8; 32]", true);` – wdyt?

Robbepop:
Have you considered simply making use of the actual type metadata in order to infer types that are wrapped in a `scale_codec::Compact` type? I mean this library literally already supports this use case so the adding of the boolean flag is redundant if we are strict.

dvdplm:
@Robbepop you mean that a field `#[scale(compact)] a:  i32,` would become `a: parity_scale_compact::Compact<i32>`? Yes, I did try that route first (as you outlined that as your preferred way in #8) but when I saw what `#[scale(compact)]` actually expands to I thought I'd try this route instead to see where it leads.

For example:

```rust
#[derive(Encode)]
pub enum MultiAddress<AccountId, AccountIndex> {
    /// It's an account ID (pubkey).
    Id(AccountId),
    /// It's an account index.
    Index(#[codec(compact)] AccountIndex),
    /// It's some arbitrary raw bytes.
    Raw(Vec<u8>),
    /// It's a 32 byte representation.
    Address32([u8; 32]),
    /// Its a 20 byte representation.
    Address20([u8; 20]),
}
```

…expands to this:

```rust
impl<AccountId, AccountIndex> _parity_scale_codec::Encode for MultiAddress<AccountId, AccountIndex>
where
    AccountId: _parity_scale_codec::Encode,
    AccountId: _parity_scale_codec::Encode,
    AccountIndex: _parity_scale_codec::HasCompact,
{
    fn encode_to<__CodecOutputEdqy: _parity_scale_codec::Output>(
        &self,
        __codec_dest_edqy: &mut __CodecOutputEdqy,
    ) {
        match *self {
            MultiAddress::Id(ref aa) => {
                __codec_dest_edqy.push_byte(0usize as u8);
                __codec_dest_edqy.push(aa);
            }
            MultiAddress::Index(ref aa) => {
                __codec_dest_edqy.push_byte(1usize as u8);
                {
                    __codec_dest_edqy.push (&<<AccountIndex as _parity_scale_codec::HasCompact>::Type as _parity_scale_codec::EncodeAsRef< '_ , AccountIndex >>::RefType::from(aa));
                }
            }
            MultiAddress::Raw(ref aa) => {
                __codec_dest_edqy.push_byte(2usize as u8);
                __codec_dest_edqy.push(aa);
            }
            MultiAddress::Address32(ref aa) => {
                __codec_dest_edqy.push_byte(3usize as u8);
                __codec_dest_edqy.push(aa);
            }
            MultiAddress::Address20(ref aa) => {
                __codec_dest_edqy.push_byte(4usize as u8);
                __codec_dest_edqy.push(aa);
            }
            _ => (),
        }
    }
}
```

src/ty/fields.rs
```diff
@@ -86,6 +86,17 @@ pub struct Field<T: Form = MetaForm> {
     ty: T::Type,
     /// The name of the type of the field as it appears in the source code.
     type_name: T::String,
+    /// This field should be encode/decoded as a
+    /// [`Compact`](parity_scale_codec::Compact) field
+    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "is_false", default))]
+    compact: bool,
```
Robbepop, line 92: I think an even more compact representation for the exact same use case would be to use yet another enum variant in order to represent compacted types. The idea is to add another variant `Compact(TypeDefCompact<T>)` to the `TypeDef` enum: https://docs.rs/scale-info/0.4.1/scale_info/enum.TypeDef.html
The `TypeDefCompact` type internally could look like this:
```rust
pub struct TypeDefCompact<T>
where
    T: Form = MetaForm
{
    #[serde(rename = "type")]
	type_param: T::Type,
}
```
Then instead of requiring each type to have this `bool` field we'd simply have a compact variant only for those type definitions that are actually compactly encoded. Since this is kind of a special case this design would suite much better. Also it would better reflect the encoding structure.

Robbepop (changes requested):

src/ty/fields.rs
ascjones: At first glance I think this solution is more elegant than having the field, while still making it a first class citizen (unlike just using a type e.g. `Compact<T>`.

dvdplm:
Closing in favour of #58

PR #44, opened 2021-01-04
Handle more SCALE attributes: skip, index

### LINKED ISSUE #43: Handle `#[codec(skip)]`
When a type is annotated with `#[codec(skip)]` the derived `TypeInfo` should skip it also.

### LINKED ISSUE #45: Handle #[codec(index)] on variants

dvdplm:
Handle `#[codec(skip)]` and `#[codec(index = $int)]` (on variants) and use the utility code copied from `parity-scale-codec-derive`.

In `parity-scale-codec-derive` there is a fair amount of validation made on attributes being in the right place. This PR does not include those checks – I assume that a type containing any of these attributes will derive `Encode` and so run the validation so duplicating them seems dumb.

~~Builds on #42~~
Closes #43 
Closes #45


.github/workflows/rust.yml
```diff
@@ -25,27 +25,30 @@ jobs:
 
     - name: fmt
       run: |
+        cargo generate-lockfile --verbose && cargo update -p funty --precise "1.1.0" --verbose
```
dvdplm, line 28: @ascjones this is a bit terrible – do you know of a better way? (Context: https://github.com/myrrlyn/funty/issues/3)

ascjones: I'd be okay with adding an explicit pin to the `Cargo.toml`, I've seen other big projects e.g. https://github.com/Geal/nom/commit/dba1e0a9937067201d7e706f7a6043834d2acfc7 just do that.

dvdplm: I did that [here](https://github.com/paritytech/parity-scale-codec/pull/250) but there is not consensus for that approach.

dvdplm: > an explicit pin to the `Cargo.toml`,

Also, there'd be three different `Cargo.toml`s to fix, so there's an argument for keeping all the ugliness here in one place.

ascjones: I'm not sure what a nice solution to this is, off the top of my head.

dvdplm: Me neither.

derive/src/utils.rs
```diff
@@ -0,0 +1,104 @@
...
+//         if let NestedMeta::Meta(Meta::Path(ref path)) = meta {
+//             if path.is_ident("compact") {
+//                 return Some(())
+//             }
+//         }
+
+//         None
+//     })
+//     .is_some()
+// }
```
ascjones, line 70: Should be removed, or the equivalent method in lib moved here

derive/src/utils.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        None
+    })
+    .is_some()
+}
+
+fn find_meta_item<'a, F, R, I>(itr: I, pred: F) -> Option<R>
+where
+    F: FnMut(&NestedMeta) -> Option<R> + Clone,
```
ascjones, line 88: I know you copied this but does it need to be `FnMut`, will `Fn` do?

derive/src/lib.rs
```diff
@@ -229,27 +228,17 @@ fn generate_composite_type(data_struct: &DataStruct, scale_info: &Ident) -> Toke
...
-            match lit_int.base10_parse::<u64>() {
-                Ok(i) => i,
-                Err(err) => return err.to_compile_error(),
+    let variants = variants
+        .into_iter()
+        .enumerate()
+        .filter(|(_, v)| !utils::should_skip(&v.attrs))
+        .map(|(i, v)| {
+            let name = &v.ident;
+            let discriminant = utils::variant_index(v, i);
```
ascjones, line 237: I believe we will need to add the `index` to regular enum variants too

ascjones (commented):
We'll need to add the `index` to regular enum variants too.

derive/src/lib.rs
dvdplm: You are correct, good catch. 
Opened https://github.com/paritytech/scale-info/issues/79 to address this in a follow-up PR.

src/tests.rs
```diff
@@ -183,3 +183,103 @@ fn basic_struct_with_phantoms() {
...
+                    "C",
+                    Fields::unnamed()
+                        .field_of::<u16>("u16")
+                        .field_of::<u32>("u32"),
+                )
+                .variant_unit("D"),
+        );
+
+    assert_type!(RustEnum, ty);
+}
```
ascjones, line 285: These tests don't add that much value, just that `type_info` returns what it returns. Perhaps good as examples of how to manually construct type info for enums, but that would be better as doc tests.

test_suite/tests/derive.rs
```diff
@@ -142,6 +142,27 @@ fn c_like_enum_derive() {
...
 
+#[test]
+fn c_like_enum_derive_with_scale_index_set() {
+    #[allow(unused)]
+    #[derive(TypeInfo, Encode)]
+    enum E {
+        A,
+        B = 10,
+        #[codec(index = 13)]
+        C,
```
ascjones, line 153: I think we need two more cases to test:
-  `D` variant, ( == `3`?) to cover the case of a variant without an explicit index after some overridden ones.
-  `E` variant with both an explicit discriminant and an `index` attribute (index overrides discriminant?)

test_suite/tests/ui/fail_with_invalid_codec_attrs.stderr
```diff
@@ -9,3 +9,11 @@ error: Invalid attribute on field, only `#[codec(skip)]`, `#[codec(compact)]` an
...
 13 |     Thing(#[codec(index = 3, compact)] u32),
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error: proc-macro derive panicked
+  --> $DIR/fail_with_invalid_codec_attrs.rs:11:10
+   |
+11 | #[derive(TypeInfo, Encode)]
+   |          ^^^^^^^^
+   |
+   = help: message: Internal error, parse_meta must have been checked: Error("expected literal")
```
ascjones, line 19: Not a very friendly error message.

derive/src/utils.rs
```diff
@@ -0,0 +1,109 @@
...
+fn find_meta_item<'a, F, R, I>(itr: I, pred: F) -> Option<R>
+where
+    F: Fn(&NestedMeta) -> Option<R> + Clone,
+    I: Iterator<Item = &'a Attribute>,
+{
+    itr.filter_map(|attr| {
+        if attr.path.is_ident("codec") {
+            if let Meta::List(ref meta_list) = attr
+                .parse_meta()
+                .expect("Internal error, parse_meta must have been checked")
```
ascjones, line 100: This is not a very helpful error, as shown in the UI test.

derive/src/utils.rs
```diff
@@ -0,0 +1,109 @@
...
+    })
+    .is_some()
+}
+
+fn find_meta_item<'a, F, R, I>(itr: I, pred: F) -> Option<R>
+where
+    F: Fn(&NestedMeta) -> Option<R> + Clone,
+    I: Iterator<Item = &'a Attribute>,
+{
+    itr.filter_map(|attr| {
```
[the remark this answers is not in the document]
dvdplm: IMO also, but `find_map` takes `&mut self` which makes things annoying/impossible. :/

src/tests.rs
dvdplm: Fair point, I added them indeed while tinkering with the api. I'll move some to doc tests.

test_suite/tests/derive.rs
dvdplm: > (index overrides discriminant?)

Yes: 1. `index` attr, 2. explicit discriminant, 3. order of appearance (aka implicit discriminant).

Will add more cases.

test_suite/tests/ui/fail_with_invalid_codec_attrs.stderr
dvdplm: Good pt, I'll improve.

src/tests.rs
dvdplm: The doc tests already covers this so removed them instead.

test_suite/tests/ui/fail_with_invalid_codec_attrs.stderr
```diff
@@ -16,4 +16,4 @@ error: proc-macro derive panicked
 11 | #[derive(TypeInfo, Encode)]
    |          ^^^^^^^^
    |
-   = help: message: Internal error, parse_meta must have been checked: Error("expected literal")
+   = help: message: Internal error. `#[codec(…)]` attributes must be checked in `parity-scale-codec`. This is a bug.: Error("expected literal")
```
ascjones, line 19: I still don't 100% get this - the errors above were generated by `parity-scale-codec` so the check appears to have happened correctly - so this error is not necessary?

dvdplm: Right, it’s there to chastise sloppy developers and should never reach users. It means that the attribute syntax checks never happened or are buggy; the derive author has to explicitly make a method call so there’s nothing stopping human error. 
We could unwrap here but that’s not better. 
The trybuild is here as a guard against buggy parity-scale-codec releases.

ascjones: >  should never reach users

Forgive me :see_no_evil: , but doesn't this trybuild demonstrate that this error does in fact reach the user even though the syntax checks did happen?

dvdplm: You are right and I am wrong. First of all I'm heaping on a bunch of different syntax errors in the same `enum` which makes it hard to understand what is going on. If I use a single-syntax error at a time it's easier to follow.

```rust
#[derive(Encode, TypeInfo)]
enum EnumAttrsValidation {
    #[codec(index = a)]
    Thing(u32),
}
```

Gives this error:

```
error: proc-macro derive panicked
  --> $DIR/fail_with_invalid_codec_attrs.rs:17:18
   |
17 | #[derive(Encode, TypeInfo)]
   |                  ^^^^^^^^
   |
   = help: message: Internal error. `#[codec(…)]` attributes must be checked in `parity-scale-codec`. This is a bug.: Error("expected literal")

error: expected literal
  --> $DIR/fail_with_invalid_codec_attrs.rs:19:21
   |
19 |     #[codec(index = a)]
   |                     ^
```

So `parity-scale-codec` is giving the proper error message back, whereas `scale-info` just panics. What I was trying to say with this test is that `scale-info` will not try to check attributes from `parity-scale-codec`. 

We have no way of knowing if the `Encode` macro ran or how it went, so we can either panic or re-implement their attribute checking and print the proper error twice.

I think we should panic but my error message must change. How about: `"Bad index in `#[codec(index = …)]`: Error("expected literal")`.

Wdyt?

ascjones: Cool, maybe could mention somehow to see the other error generated by `parity-scale-codec`.

ascjones (approved):
LGTM!

dvdplm:
@Robbepop This one should be pretty non-controversial.

PR #47, opened 2021-01-06
Fix type name scrubbing to handle nested tuples

dvdplm:


ascjones (commented):
Could we add a simple test case for this in https://github.com/paritytech/scale-info/blob/master/test_suite/tests/derive.rs?

ascjones (approved):
LGTM

PR #48, opened 2021-01-07
Add a new TypeDef variant to handle PhantomData

### LINKED ISSUE #7: Remove PhantomData

### LINKED ISSUE #8: Support Compact types

dvdplm:
See https://github.com/paritytech/scale-info/pull/31#issuecomment-754722413 for the background.

Supersedes #31
Closes #7 
﻿
Note that we do not yet have support for `#[codec(skip)]` (implemented in PR #44) so when that lands we'll want to amend the tests here (or there, if this lands first).

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -245,7 +245,7 @@ impl<T> TypeInfo for PhantomData<T>
 where
     T: TypeInfo + ?Sized + 'static,
 {
-    type Identity = T;
+    type Identity = Self;
 
     fn type_info() -> Type {
         TypeDefPhantom::new(MetaType::new::<T>()).into()
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -17,17 +17,16 @@ use crate::prelude::{
     collections::BTreeMap,
     marker::PhantomData,
     string::String,
-    vec,
     vec::Vec,
 };
 
 use crate::{
     build::*,
-    meta_type,
     MetaType,
     Path,
     Type,
     TypeDefArray,
+    TypeDefPhantom,
     TypeDefPrimitive,
     TypeDefSequence,
     TypeDefTuple,
@@ -240,12 +239,9 @@ impl<T> TypeInfo for PhantomData<T>
 where
     T: TypeInfo + ?Sized + 'static,
 {
-    type Identity = Self;
+    type Identity = T;
 
     fn type_info() -> Type {
-        Type::builder()
-            .path(Path::prelude("PhantomData"))
-            .type_params(vec![meta_type::<T>()])
-            .composite(Fields::unit())
+        TypeDefPhantom::new(MetaType::new::<T>()).into()
     }
 }
--- a/src/ty/mod.rs
+++ b/src/ty/mod.rs
@@ -410,6 +410,15 @@ where
 }
 
 /// A type describing a `PhantomData<T>` type.
+///
+/// In the context of SCALE encoded types, including `PhantomData<T>` types in
+/// the type info  might seem surprising. The reason to include this information
+/// is that there could be situations where it's useful and because removing
+/// `PhantomData` items from the derive input quickly becomes a messy
+/// syntax-level hack (see PR https://github.com/paritytech/scale-info/pull/31).
+/// Instead we take the same approach as `parity-scale-codec` where users are
+/// required to explicitly skip fields that cannot be represented in SCALE
+/// encoding, using the `#[codec(skip)]` attribute.
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(
     feature = "serde",
--- a/src/ty/mod.rs
+++ b/src/ty/mod.rs
@@ -414,8 +414,8 @@ where
 #[cfg_attr(
     feature = "serde",
     serde(bound(
-        serialize = "T::Type: Serialize, T::String: Serialize",
-        deserialize = "T::Type: DeserializeOwned, T::String: DeserializeOwned",
+        serialize = "T::Type: Serialize",
+        deserialize = "T::Type: DeserializeOwned",
     ))
 )]
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Decode, Debug)]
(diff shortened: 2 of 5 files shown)


test_suite/tests/json.rs
```diff
@@ -197,6 +196,30 @@ fn test_struct() {
...
+
+    assert_json_for_type::<Struct<u8>>(json!({
+        "path": ["json", "Struct"],
+        "params": [1],
+        "def": {
+            "composite": {
+                "fields": [
+                    // TODO: dp – I was expecting "a" to be type 1 and "b" to be type 2; what am I missing?
+                    { "name": "a", "type": 2, "typeName": "i32" },
+                    { "name": "b", "type": 1, "typeName": "PhantomData<T>" },
```
dvdplm, line 216: @ascjones I was a little surprised by the values for `"type"` here. Is it expected that the type parameter goes first into the registry, as type 1? Or am I missing something?

Robbepop: Encoding is a bit complicated due to more exotic use cases such as recursive types.

dvdplm: So not a source for concern then?

Robbepop: As long as the encoding is deterministic everything is alright. Exact ordering does not matter as long as it is consistent.

test_suite/tests/derive.rs
```diff
@@ -76,7 +76,6 @@ fn struct_derive() {
     assert_type!(SelfTyped, self_typed_type);
 }
 
-// TODO: dp – no need to keep all of these tests; pick one and toss the rest.
```
Robbepop: Why is there no need for the other tests? As it looks on the surface level they are all testing the feature under different scenario.

dvdplm: I copied the tests from the other PR where there was a greater need for validating that no `PhantomData` items were allowed in different scenarios. 
In this PR all `PhantomData` members are allowed through and I think it's a bit of an overkill to assert that yes, Phantoms are allowed everywhere (unless skipped, which isn't available in this PR).

src/impls.rs
```diff
@@ -240,12 +239,9 @@ impl<T> TypeInfo for PhantomData<T>
 where
     T: TypeInfo + ?Sized + 'static,
 {
-    type Identity = Self;
+    type Identity = T;
```
ascjones, line 242: I think this is wrong, should be `Self`. Otherwise `PhantomData<bool>` == `bool`

test_suite/tests/derive.rs
```diff
@@ -73,6 +76,78 @@ fn struct_derive() {
...
+        m: PhantomData<T>,
+    }
+
+    let ty = Type::builder()
+        .path(Path::new("P", "derive"))
+        .type_params(tuple_meta_type!(bool))
+        .composite(
+            Fields::named()
+                .field_of::<u8>("a", "u8")
+                .field_of::<bool>("m", "PhantomData<T>"),
```
ascjones, line 94:
```suggestion
                .field_of::<PhantomData<bool>>("m", "PhantomData<T>"),
```

ascjones (changes requested):
Just need to fix the identity of `PhantomData<T>`, some tests may fail

src/ty/mod.rs
```diff
@@ -118,6 +118,16 @@ impl From<TypeDefTuple> for Type {
     }
 }
 
+impl From<TypeDefPhantom> for Type {
+    fn from(phantom: TypeDefPhantom) -> Self {
+        Self::new(
+            Path::prelude("PhantomData"),
```
ascjones, line 124: Arguably the name is redundant here, could just be `voldemort`.

src/ty/mod.rs
```diff
@@ -118,6 +118,16 @@ impl From<TypeDefTuple> for Type {
...
+            vec![phantom.type_param],
```
ascjones, line 125: This is also arguably redundant since the type is already part of the definition, but it does make it consistent with e.g. `TypeDefArray` and `TypeDefTuple`

test_suite/tests/derive.rs
```diff
@@ -73,6 +76,78 @@ fn struct_derive() {
...
+            Fields::named()
+                .field_of::<u8>("a", "u8")
+                .field_of::<PhantomData<bool>>("m", "PhantomData<T>"),
+        );
+
+    assert_type!(P<bool>, ty);
+}
+
+#[test]
+fn nested_phantom_types_are_derived_in_structs() {
```
ascjones, line 101: What is this actually testing in respect to the implementation here?

I'd argue that this is a core feature of the library, supporting nested custom types, independent of the `PhantomData` feature.

ascjones (approved):
LGTM

Robbepop (commented):
All in all code looks good.
The only thing I am missing is an in-depth comment explaining the design rationals behind this PR so that we won't forget why we chose this design instead of others we previously talked about. I always love finding such comments when I forgot about certain rationals after months of not looking at the issue.

test_suite/tests/derive.rs
dvdplm: Yeah I don't disagree. It's redundant.

src/ty/mod.rs
dvdplm: No it's not consistent: all the others are `Self::new(Path::voldemort(), Vec::new(), $THING)`.

I did it this way to get what seemed like richer metadata when serializing, i.e. this:

```json
        "path": ["PhantomData"],
        "def": { "phantom": { "type": 1 } },
        "params": [1]
```

```json
        "def": { "phantom": { "type": 1 } },
```

…so I'll change this as suggested.

dvdplm:
> The only thing I am missing is an in-depth comment explaining the design rationals behind this PR

@Robbepop added https://github.com/paritytech/scale-info/pull/48/commits/f259e70756893f41738d7925748c24ef183006d2 – happy to tweak that in a separate PR if you want.

Robbepop:
I'd love to have the example of reconstructing rust types given the scale info metadata as motivation for having PhantomData deginitions.

PR #49, opened 2021-01-11
Remove unused function `MetaType::of()`

dvdplm:
More of a question really. Just noticed this function is not used anywhere and not tested. Should it be removed perhaps? Otherwise I'll add a test for it.﻿

### DIFF
--- a/src/meta_type.rs
+++ b/src/meta_type.rs
@@ -96,14 +96,6 @@ impl MetaType {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Robbepop:
I think i added it a long time ago as convenient function for the proc macros. If it is really unused and deemed not useful enough we can as well just remove it.

ascjones (approved):
I can't think of a case where we would use a reference to construct a `MetaType`, so fine for me to remove.

PR #52, opened 2021-01-13
Derive Decode for PortableForm

dvdplm:
While working on [`frame-metadata`](paritytech/frame-metadata#3) I realized that we might need `Decode` for `PortableForm` as well.

E.g.:

```rust
pub struct RuntimeMetadataV13<S: FormString = &'static str> {
	pub types: PortableRegistry<S>,
	/// Metadata of all the modules.
	pub modules: Vec<ModuleMetadata<PortableForm>>,
	/// Metadata of the extrinsic.
	pub extrinsic: ExtrinsicMetadata<PortableForm>,
}
```

### DIFF
--- a/src/form.rs
+++ b/src/form.rs
@@ -41,6 +41,8 @@ use crate::{
     meta_type::MetaType,
 };
 
+use scale::Decode;
+
 #[cfg(feature = "serde")]
 use serde::Serialize;
 
@@ -88,7 +90,7 @@ impl Form for MetaForm {
 ///
 /// `type String` is owned in order to enable decoding
 #[cfg_attr(feature = "serde", derive(Serialize))]
-#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug)]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug, Decode)]
 pub struct PortableForm<S = &'static str>(PhantomData<S>);
 
 impl<S> Form for PortableForm<S>


ascjones (commented):
Why does it need `Decode`? No actual instances of this type.

I can already decode `PortableRegistry` without this.

dvdplm:
> Why does it need `Decode`? No actual instances of this type.

My compiler error was this:

```
error[E0277]: the trait bound `Vec<ModuleMetadata<PortableForm>>: Decode` is not satisfied
   --> frame-metadata/src/v13.rs:60:2
    |
60  | /     /// Metadata of all the modules.
61  | |     pub modules: Vec<ModuleMetadata<PortableForm>>,
    | |__________________________________________________^ the trait `Decode` is not implemented for `Vec<ModuleMetadata<PortableForm>>`
    | 
   ::: /Users/aggron/.cargo/registry/src/github.com-1ecc6299db9ec823/parity-scale-codec-1.3.6/src/codec.rs:311:18
    |
311 |       fn decode<I: Input>(value: &mut I) -> Result<Self, Error>;
    |                    ----- required by this bound in `_::_parity_scale_codec::Decode::decode`
    |
    = help: the following implementations were found:
              <Vec<T> as Decode>
```

(the code isn't committed yet, but based off tomeks work on frame-metadata)

ascjones:
I can see the problem, you need to pass the `S` to `PortableForm` e.g. `ModuleMetadata<PortableForm<S>>`

Check out my [branch](https://github.com/paritytech/frame-metadata/blob/b1008c8f6032eb2ed27a590b28ac39356b7b480f/frame-metadata/src/v13.rs#L57) which defines exactly that struct. This is the branch I am currently using to convert substrate to use v13.

PR #53, opened 2021-01-18
Add a TypeDef to handle Compact types

### LINKED ISSUE #8: Support Compact types

dvdplm:
Update the derive macro to handle the `#[codec(compact)` attribute from `parity-scale-codec` using a new `TypeDefCompact` variant similarly to #48.

Closes #8 

Supersedes #42


test_suite/tests/derive.rs
```diff
@@ -233,6 +235,69 @@ fn associated_types_derive_without_bounds() {
...
+        );
+
+    assert_type!(Dense, ty);
+
+    // TODO: These two tests both pass. This illustrates a doubt I have that
+    // this is the right way to go. With this PR we take `#[codec(compact)]` to
+    // mean that the type is encoded/decoded as `Compact<T>` but in reality the
+    // output of the `Encode` macro is something quite different and not
+    // `Compact<T>`.
+    // Maybe this is fine, maybe it's not, just not sure.
```
dvdplm, line 263: @ascjones I left this TODO as an illustration of the crux of the PR: we will take `#[codec(compact)] thing: T` to mean `thing: Compact<T>`. I am not 100% positive this is correct so I wanted to point it out.

ascjones: I think it is okay, since I believe both these will result in the exact same encoded representation (@thiolliere?), which is what we are describing here. 

I presume one should usually prefer the attribute rather than the type for fields because that is more ergonomic for usage of the type, we shouldn't care about how it is encoded when using it.

src/prelude.rs
```diff
@@ -57,3 +57,5 @@ cfg_if! {
         };
     }
 }
+
+pub use scale::Compact;
```
ascjones, line 61: Why reexport `Compact`?

src/ty/mod.rs
```diff
@@ -405,6 +397,49 @@ where
     }
 }
 
+/// A type wrapped in [`Compact`].
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Decode, Debug)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(
+    feature = "serde",
+    serde(bound(
+        // TODO: check these bounds
```
ascjones, line 406: Actually I think they are unnecessary because `T::Type` is used directly as a field type and `serde` must add the bounds automatically in that case. Just made https://github.com/paritytech/scale-info/pull/60 which removes other redundant ones.

test_suite/tests/derive.rs
```diff
@@ -11,16 +11,18 @@
...
+#![cfg_attr(not(feature = "std"), no_std)]
 
 use pretty_assertions::assert_eq;
+use scale::Encode;
 use scale_info::{
     build::*,
+    prelude::{
+        boxed::Box,
+        marker::PhantomData,
+        Compact,
```
ascjones, line 24: I'm not sure we do want `codec::Compact` in our prelude. I think we already have a dependency on `parity_scale_codec` in this test crate which we can use

test_suite/tests/derive.rs
thiolliere: yes, the macro will use the type: `<$FIELD as HasCompact>::Type`
The trait HasCompact is implemented automatically for `T` and associated type to be `Compact<T>`.
But actually a user of parity-scale-codec can implemement HasCompact himself on its type and have something different that `Compact<T>` used when encoding/decoding.

One solution could be to Seal the trait in parity-scale-codec in order to prevent user to do that.

But I don't know about the architecture of scale-info, but it is possible to actually use the correct type: `<$FIELD as HasCompact>::Type` ? (which will then resolve to `Compact<T>` most of the time)

src/prelude.rs
dvdplm: The idea was to put `Compact` on the same footing as the other re-exported types in the `prelude`, mostly as a marker that this is a type that `scale-info` expects and respects I guess. What is the point of adding a `use scale::Compact` here?

src/ty/mod.rs
dvdplm: Nice, good riddance.

test_suite/tests/derive.rs
dvdplm: See comment above, happy to remove it. The point was more about signalling than a technical need.

derive/src/lib.rs
```diff
@@ -126,20 +129,37 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
...
         .collect()
 }
 
+/// Look for a `#[codec(compact)]` outer attribute.
+fn is_compact(f: &Field) -> bool {
+    f.attrs.iter().any(|attr| {
+        let mut is_compact = false;
+        if attr.style == AttrStyle::Outer && attr.path.is_ident("codec") {
+            if let Ok(Meta::List(MetaList { nested, .. })) = attr.parse_meta() {
+                if let Some(NestedMeta::Meta(Meta::Path(path))) = nested.iter().next() {
```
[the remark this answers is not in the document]
dvdplm: Good catch, let's do that in the next PR though (i.e. #44 and #45) as we don't really have `skip` here yet.

src/prelude.rs
Robbepop: I think his point is about simply not re-exporting it and expecting the user to instead use `Compact` directly from the `parity-scale-codec` crate.

derive/src/lib.rs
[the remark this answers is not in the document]
ascjones: I think it should be in this PR because if `compact` is not in the first position it will be ignored, because AFAICT only the first one is checked via `nested.iter().next()`

src/prelude.rs
ascjones: > The idea was to put Compact on the same footing as the other re-exported types in the prelude

The purpose of the `prelude` is to provide a common interface for required types between `std` and `no_std` environments. Perhaps `prelude` is not a great name then.

Robbepop: Yes the name really is bad (sorry for introducing it :sweat_smile: ) . A better name would be something like `alloc_compat` because that's what it actually is since it is a compatibility layer between `std` and `alloc` Rust crates.

test_suite/tests/derive.rs
dvdplm: Hmm. I think I'm running into the limits of my brain power here. When I add a `HasCompact` bound [here](https://github.com/paritytech/scale-info/blob/5649a666bb3e8228ebdf59e5844ddf22604594a5/src/ty/fields.rs#L136) and try to `MetaType::new::<<T as HasCompact>::Type>()` I end up with infinite recursion:

```
error[E0275]: overflow evaluating the requirement `Compact<_>: Decode`
  |
  = help: consider adding a `#![recursion_limit="256"]` attribute to your crate (`scale_info`)
  = note: required because of the requirements on the impl of `Decode` for `Compact<_>`
  = note: 126 redundant requirements hidden
  = note: required because of the requirements on the impl of `Decode` for `Compact<_>`
  ```

(it's infinite; the error message increases the suggestion indefinitely, 256, 512, …, … 8192 etc)

Comment (ascjones) [reply test_suite/tests/derive.rs]: I guess `<T as HasCompact>::Type` must be `Self`?

Comment (ascjones) [reply test_suite/tests/derive.rs]: Or some other way ends up calling back into this `TypeInfo` impl

Comment (thiolliere) [reply test_suite/tests/derive.rs]: this compiles to me:
```diff
diff --git a/src/build.rs b/src/build.rs
index 9669aee..d288320 100644
--- a/src/build.rs
+++ b/src/build.rs
@@ -263,7 +263,8 @@ impl FieldsBuilder<NamedFields> {
     /// Add a named, [`Compact`] field of type `T`.
     pub fn compact_of<T>(mut self, name: &'static str, type_name: &'static str) -> Self
     where
-        T: TypeInfo + 'static,
+               T: scale::HasCompact,
+               <T as scale::HasCompact>::Type: TypeInfo + 'static,
     {
         self.fields
             .push(Field::compact_of::<T>(Some(name), type_name));
@@ -284,7 +285,8 @@ impl FieldsBuilder<UnnamedFields> {
     /// Add an unnamed, [`Compact`] field of type `T`.
     pub fn compact_of<T>(mut self, type_name: &'static str) -> Self
     where
-        T: TypeInfo + 'static,
+               T: scale::HasCompact,
+               <T as scale::HasCompact>::Type: TypeInfo + 'static,
     {
         self.fields.push(Field::compact_of::<T>(None, type_name));
         self
diff --git a/src/ty/fields.rs b/src/ty/fields.rs
index 9229ab9..974fa09 100644
--- a/src/ty/fields.rs
+++ b/src/ty/fields.rs
@@ -133,9 +133,10 @@ impl Field {
     /// Creates a new [`Compact`] field.
     pub fn compact_of<T>(name: Option<&'static str>, type_name: &'static str) -> Field
     where
-        T: TypeInfo + 'static,
+               T: scale::HasCompact,
+               <T as scale::HasCompact>::Type: TypeInfo + 'static,
     {
-        Self::new(name, MetaType::new::<scale::Compact<T>>(), type_name)
+        Self::new(name, MetaType::new::<<T as scale::HasCompact>::Type>(), type_name)
     }
 }
 
```

I see rustc fails to give an ok error message when src/build bounds are not changed

dvdplm: > I see rustc fails to give an ok error message when src/build bounds are not changed

Yeah was wondering about that, it didn't complain. :/ Thank you though, the above type checks for me, but this code still causes infinite recursion:

```rust
Fields::named().compact_of::<u16>("b", "u16"),
```
Full test [here](https://github.com/paritytech/scale-info/blob/5649a666bb3e8228ebdf59e5844ddf22604594a5/test_suite/tests/derive.rs#L261-L284), run with `cargo t --all`.

@ascjones do you mean `type Identity = <Self as HasCompact>::Type;` [here](https://github.com/paritytech/scale-info/blob/5649a666bb3e8228ebdf59e5844ddf22604594a5/src/impls.rs#L260)? In itself it doesn't compile. :/

thiolliere: I think it is because derive macro doesn't bound the correct bound on the compact type. (see DM)

dvdplm: This is now sorted thanks to @thiolliere. 💯

src/prelude.rs
dvdplm: > we could even merge both crates in the future. haven't thought this through yet, though.

I've thought about this too. They are quite entangled. But that's a future concern I'd say.

Removed the re-export.

derive/src/lib.rs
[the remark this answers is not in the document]
dvdplm: Fair enough. On it.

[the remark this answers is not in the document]
ascjones: I'm sure it can wait until Monday morning though :wine_glass:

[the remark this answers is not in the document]
dvdplm: On a second look I think you're actually not reading in the right place. Attribute validation happens [here](https://github.com/paritytech/parity-scale-codec/blob/5593ee726f3f7b03ef20e1bb9e3ae8b9ef029b9e/derive/src/lib.rs#L139), which then calls `check_field_attribute()`, which [enforces a single item](https://github.com/paritytech/parity-scale-codec/blob/5593ee726f3f7b03ef20e1bb9e3ae8b9ef029b9e/derive/src/utils.rs#L202) (the variant equivalent is [here](https://github.com/paritytech/parity-scale-codec/blob/5593ee726f3f7b03ef20e1bb9e3ae8b9ef029b9e/derive/src/utils.rs#L233)).

I'm not sure which attribute combinations would make sense for `parity-scale-codec` to support. `skip` + `compact` not so much, and I'm not sure `compact` + `encoded_as` either. Maybe `index` + `compact`/`encoded_as`? But so far I think they have been pretty strict about attributes as the macro [may panic](https://github.com/paritytech/parity-scale-codec/blob/master/derive/src/utils.rs#L146-L148) otherwise.

The `find_meta_item()` utility function you linked to is general purpose (and totally neat, so I'm cribbing it in #44).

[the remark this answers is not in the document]
ascjones: You are right it will not allow multiple attributes: https://github.com/paritytech/parity-scale-codec/blob/master/derive/src/utils.rs#L201.

This crosses over into the discussion about whether/how `scale-info` and `parity-scale-codec` can be combined, because we are relying on this attribute validation. So we only become vulnerable here (i.e, the attribute could be ignored) if this validation is relaxed to allow some future valid combination of attributes (I can see now that `#[codec(skip, compact)]` is wrong).

So could either make it future compatible now, or just add a compile test which checks that this validation is in place. Don't mind if you do it here or in one of the follow ups.

ascjones (approved):
Awesome! Would like a :heavy_check_mark: from @Robbepop before merging.

derive/src/trait_bounds.rs
```diff
@@ -51,10 +51,21 @@ pub fn make_where_clause<'a>(
...
+    types.into_iter().for_each(|(ty, is_compact)| {
+        // Compact types need extra bounds, T: HasCompact and <T as
+        // HasCompact>::Type: TypeInfo + 'static
+        if is_compact {
+            where_clause
+                .predicates
+                .push(parse_quote!(#ty : ::scale::HasCompact));
+            where_clause
+                .predicates
+                .push(parse_quote!(<#ty as ::scale::HasCompact>::Type : ::scale_info::TypeInfo + 'static));
```
Robbepop, line 63: I think our use of `::scale` is becoming a problem. ink! imports `parity-scale-codec` crate as `scale` but from what I remember Substrate imports it as `codec`. We need a solution that works for both. Probably Basti's https://crates.io/crates/proc-macro-crate could help here. For now I am fine and we can fix this later.

Robbepop (approved):
LGTM

One nit pick that should be handled as a follow-up PR.

derive/src/lib.rs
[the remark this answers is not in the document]
thiolliere: Yes for now it seems all attribute are exclusive:

For field attribute I think they are all exclusive: either compact, or skip or encoded_as or nothing.
For variant attribute they are also exclusive: either skip or index or nothing.

derive/src/trait_bounds.rs
ordian: this also seem to break primitive-types: https://github.com/paritytech/parity-common/pull/519

PR #54, opened 2021-01-22
Use const generics for array TypeInfo impls

ascjones:
Waiting for https://github.com/rust-lang/rust/issues/74878 to make it into stable rust. Current status: stabilized on nightly.


Robbepop (approved):
I like

PR #55, opened 2021-01-26
Update codec to 2.0 and release new version (before primitive-types release a new version for codec 2.0)

thiolliere:
I'm not sure about this version bump:

substrate doesn't depend on scale-info, but primitive-types does depend optionally on scale-info, and primitive-types will get soon upgraded to 2.0 thus I think it is better if it also update scale-info to 0.5.

Apart from this, it seems we don't need to bump the derive crate, but I'm not sure, maybe it is better not to mix different version of scale-codec when deriving scale-info.
you can take a look at the changelog to make your mind https://github.com/paritytech/parity-scale-codec/blob/master/CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Version 0.5.0 (2021-01-27)
+
+* Bump `parity-scale-codec` from 1.0 to 2.0 [(#55)](https://github.com/paritytech/scale-info/pull/55)
+
 # Version 0.4.1 (2020-10-11)
 
 * Add missing `extern crate proc_macro;` [(#22)](https://github.com/paritytech/scale-info/pull/24)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 
 [dependencies]
 cfg-if = "1.0"
-scale-info-derive = { version = "0.2.1", path = "derive", default-features = false, optional = true }
+scale-info-derive = { version = "0.3.0", path = "derive", default-features = false, optional = true }
 serde = { version = "1", default-features = false, optional = true, features = ["derive", "alloc"] }
 derive_more = { version = "0.99.1", default-features = false, features = ["from"] }
 scale = { package = "parity-scale-codec", version = "2.0", default-features = false, features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scale-info"
-version = "0.4.1"
+version = "0.5.0"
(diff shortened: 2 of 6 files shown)


thiolliere:
I can't understand why CI is failing though

thiolliere:
I didn't look about other deps that could be bumped, I guess we do want to include 
https://github.com/paritytech/parity-common/pull/509/files (actually not sure about tokio)
and
https://github.com/paritytech/parity-common/pull/506/files

dvdplm:
> I can't understand why CI is failing though

I think it's because `rustc` is **really** old (see https://github.com/paritytech/scale-info/blob/7afea1d5928b0c0ab0299079cb54c6c1599d26a2/.github/workflows/rust.yml#L46).

ascjones:
> substrate doesn't depend on scale-info, but primitive-types does depend optionally on scale-info, and primitive-types will get soon upgraded to 2.0 thus I think it is better if it also update scale-info to 0.5.

Yes we should do a release here, though I'd like to get https://github.com/paritytech/scale-info/pull/53 included so I'll review that now.

> Apart from this, it seems we don't need to bump the derive crate

I believe we will need to for the release because there are definitely changes in there since the last release e.g. https://github.com/paritytech/scale-info/pull/30

dvdplm:
> I think it's because `rustc` is **really** old

No, I'm wrong (well, it **is** really old), I think you forgot to update [this](https://github.com/paritytech/scale-info/blob/7afea1d5928b0c0ab0299079cb54c6c1599d26a2/test_suite/derive_tests_no_std/Cargo.toml#L14).

ascjones:
> I think it's because rustc is really old

https://github.com/paritytech/scale-info/pull/56

thiolliere:
> > I think it's because `rustc` is **really** old
> 
> No, I'm wrong (well, it **is** really old), I think you forgot to update [this](https://github.com/paritytech/scale-info/blob/7afea1d5928b0c0ab0299079cb54c6c1599d26a2/test_suite/derive_tests_no_std/Cargo.toml#L14).

Yes CI is successful now

ascjones (approved):
LGTM. We can bump the derive crate as part of a PR which updates the CHANGELOG

dvdplm (approved):

thiolliere:
> LGTM. We can bump the derive crate as part of a PR which updates the CHANGELOG

Do you mean you prefer to bump derive crate ? then I can do here it is fine

dvdplm:
I'd be in favour of bumping it here, but either way is fine.

ascjones:
> > LGTM. We can bump the derive crate as part of a PR which updates the CHANGELOG
> 
> Do you mean you prefer to bump derive crate ? then I can do here it is fine

Yes please

thiolliere:
Yes done

PR #56, opened 2021-01-27
CI: use latest nightly, refactor derive trait bounds

ascjones:
- uses latest nightly with the required components
- includes a refactoring of `derive/trait_bounds` in order to make clippy happy :paperclip:

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -70,13 +70,13 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
     let mut ast: DeriveInput = syn::parse2(input.clone())?;
 
     let ident = &ast.ident;
-    trait_bounds::add(ident, &mut ast.generics, &ast.data)?;
 
     ast.generics
         .lifetimes_mut()
         .for_each(|l| *l = parse_quote!('static));
 
-    let (_, ty_generics, where_clause) = ast.generics.split_for_impl();
+    let (ty_generics, where_clause) = trait_bounds::add(ident, &ast.generics, &ast.data)?;
+
     let generic_type_ids = ast.generics.type_params().map(|ty| {
         let ty_ident = &ty.ident;
         quote! {
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -217,7 +217,10 @@ fn is_c_like_enum(variants: &VariantList) -> bool {
     // any variant has an explicit discriminant
     variants.iter().any(|v| v.discriminant.is_some()) ||
         // all variants are unit
-        variants.iter().all(|v| matches!(v.fields, Fields::Unit))
+        variants.iter().all(|v| match v.fields {
+            Fields::Unit => true,
+            _ => false,
+        })
 }
 
 fn generate_variant_type(data_enum: &DataEnum) -> TokenStream2 {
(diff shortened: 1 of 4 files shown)


dvdplm:
I think Clippy is wrong-ish here, or rather I remember trying to sort out those warnings and failing to come up with something better.

ascjones:
> I think Clippy is wrong-ish here, or rather I remember trying to sort out those warnings and failing to come up with something better.

I think it was on to something

ascjones:
> > I think Clippy is wrong-ish here, or rather I remember trying to sort out those warnings and failing to come up with something better.
> 
> I think it was on to something

Oh maybe not now I have made it happy but broken the implementation - now I remember

ascjones:
Just for my own entertainment I made a refactoring that satisfied clippy. Let me know what you think @dvdplm

derive/src/lib.rs
```diff
@@ -70,13 +70,13 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
...
 
     let ident = &ast.ident;
-    trait_bounds::add(ident, &mut ast.generics, &ast.data)?;
 
     ast.generics
         .lifetimes_mut()
         .for_each(|l| *l = parse_quote!('static));
 
-    let (_, ty_generics, where_clause) = ast.generics.split_for_impl();
+    let (ty_generics, where_clause) = trait_bounds::add(ident, &ast.generics, &ast.data)?;
```
dvdplm, line 78: This is where it looks a bit odd, we call `add()` and get back more stuff than I'd expect just from reading the method name and arguments. It's fine, better than before even, but just looks a bit odd.
I wonder what this would look like if we just went against orthodoxy and in-lined the whole `trait_bounds::add` function here directly. It would be long, yes, but perhaps more straight forward to read at the end of the day?

derive/src/trait_bounds.rs
```diff
@@ -22,41 +22,54 @@ use syn::{
...
-pub fn add(input_ident: &Ident, generics: &mut Generics, data: &syn::Data) -> Result<()> {
-    let ty_params_ids = generics
-        .type_params()
+pub fn add<'a>(
+    input_ident: &'a Ident,
+    generics: &'a Generics,
+    data: &'a syn::Data,
+) -> Result<(TypeGenerics<'a>, WhereClause)> {
+    let (_, ty_generics, where_clause) = generics.split_for_impl();
+    let mut where_clause = where_clause.cloned().unwrap_or_else(|| {
```
dvdplm, line 38: I don't remember exactly what the problem was but I think I had something similar. I found no way around having to clone things either.

dvdplm (approved):

derive/src/lib.rs
ascjones: Yes it does look odd now I have a fresh look. I have pushed some further refactoring to hopefully make it a bit more understandable.

derive/src/trait_bounds.rs
ascjones: They main thing was not using the built in `generics.make_where_clause()` which requires `&mut self`, and just copying what it does without requiring an exclusive reference.

derive/src/lib.rs
```diff
@@ -75,7 +75,8 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
         .lifetimes_mut()
         .for_each(|l| *l = parse_quote!('static));
 
-    let (ty_generics, where_clause) = trait_bounds::add(ident, &ast.generics, &ast.data)?;
+    let (_, ty_generics, _) = ast.generics.split_for_impl();
+    let where_clause = trait_bounds::make_where_clause(ident, &ast.generics, &ast.data)?;
```
dvdplm, line 79: Nice, I like it.

Issue #58, opened 2021-01-28
Reconsider handling of parameterized strings


ascjones:
Further to https://github.com/paritytech/scale-info/pull/35, we should explore ways to either improve this API or even completely change approach.

The goal is to avoid the confusing API where users need to supply `String` as a parameter to `PortableForm` when decoding/deserializing. 

Note that this is a constraint introduced by the requirement of integrating `scale-info` into substrate runtimes.

Ideas discussed with @Robbepop 

- Add a crate feature which defines the type of the `String` at compile time.
- Remove the metadata definitions from the runtime wasm in substrate and generate it as a separate process similar to `ink!`. Note that this is something I will explore anyway, and if we do it we will be able to generate metadata in the `std` environment which could just use `String`s

ascjones:
Closed by #59

PR #59, opened 2021-01-29
Add feature for enabling decoding

ascjones:
Possible solution for #58, reverting #35

If the `decode` or `std` feature is enabled then `PortableForm::String` will be an owned `String` which allow decoding. Otherwise it will be a `&'static str` (this is required for substrate runtimes).

## todo

- [x] test it out with substrate
- [x] test it out with `ink!`

### DIFF
--- a/src/form.rs
+++ b/src/form.rs
@@ -32,7 +32,6 @@
 use crate::prelude::{
     any::TypeId,
     fmt::Debug,
-    string::String,
 };
 
 use crate::{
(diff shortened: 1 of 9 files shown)


Robbepop:
>  test it out with ink!

not sure there is a need for it, is there?

ascjones:
> > test it out with ink!
> 
> not sure there is a need for it, is there?

Just wanted to make absolutely sure before merging and cutting a release. See https://github.com/paritytech/ink/pull/672

src/form.rs
```diff
@@ -53,18 +51,9 @@ pub trait Form {
...
 
-/// Trait for types which can be used to represent strings in type definitions.
-pub trait FormString:
-    AsRef<str> + PartialEq + Eq + PartialOrd + Ord + Clone + Debug
-{
-}
-
-impl FormString for &'static str {}
-impl FormString for String {}
-
```
Robbepop, line 67: glad it is gone

Robbepop (approved):
thanks for the PR! this is definitely a big improvement over the previous design.

PR #60, opened 2021-01-29
Remove redundant serde bounds

ascjones:
Where the associated type is used directly in a field e.g. `a: T::Type`, serde will add its own bounds in this case (I assume). In any case if the compiler is happy then so am I.


Robbepop:
this is neat if it preserves semantics

dvdplm (approved):
If it works it works.

Issue #65, opened 2021-02-03
Compact trait bounds overflow


ascjones:
When integrating the latest `scale-info` `master` into substrate (see https://github.com/paritytech/substrate/compare/aj-metadata-vnext)  I receive the following error:

![image](https://user-images.githubusercontent.com/75586/106740881-e5749780-6612-11eb-97c1-729608133770.png)

It appears to have been introduced in #53.

Removing the `TypeInfo` derive on `MultiAddress` makes it compile:

```diff
diff --git a/primitives/runtime/src/multiaddress.rs b/primitives/runtime/src/multiaddress.rs
index a47201008..d09cd7aca 100644
--- a/primitives/runtime/src/multiaddress.rs
+++ b/primitives/runtime/src/multiaddress.rs
@@ -21,7 +21,7 @@ use codec::{Encode, Decode};
 use sp_std::vec::Vec;
 
 /// A multi-format address wrapper for on-chain accounts.
-#[derive(Encode, Decode, PartialEq, Eq, Clone, crate::RuntimeDebug, scale_info::TypeInfo)]
+#[derive(Encode, Decode, PartialEq, Eq, Clone, crate::RuntimeDebug)]
 #[cfg_attr(feature = "std", derive(Hash))]
 pub enum MultiAddress<AccountId, AccountIndex> {
 	/// It's an account ID (pubkey).
diff --git a/primitives/runtime/src/traits.rs b/primitives/runtime/src/traits.rs
index 00fae6fe6..0fdf6d55b 100644
--- a/primitives/runtime/src/traits.rs
+++ b/primitives/runtime/src/traits.rs
@@ -203,7 +203,7 @@ pub trait Lookup {
 /// context.
 pub trait StaticLookup {
 	/// Type to lookup from.
-	type Source: Codec + Clone + PartialEq + Debug + scale_info::TypeInfo;
+	type Source: Codec + Clone + PartialEq + Debug;
 	/// Type to lookup into.
 	type Target;
 	/// Attempt a lookup.
```

Here is the expanded derived TypeInfo impl for `MultiAddress`:

```
impl<AccountId, AccountIndex> ::scale_info::TypeInfo for MultiAddress<AccountId, AccountIndex>
        where
            AccountId: ::scale_info::TypeInfo + 'static,
            AccountIndex: ::codec::HasCompact,
            <AccountIndex as ::codec::HasCompact>::Type: ::scale_info::TypeInfo + 'static,
[...]

PR #66, opened 2021-02-04
Minimal repro of type recursion overflow with compact 

ascjones:
Just to provide a minimal repro for #65, not necessarily to be merged, just to provide a playground for collaboration.

Still need to understand exectly what is going on.


test_suite/tests/ui/pass_compact_generic_with_unbounded_trait_impl.rs
```diff
@@ -0,0 +1,33 @@
...
+    type Type: TypeInfo + 'static;
+}
+
+#[derive(TypeInfo, Encode, Decode)]
+struct Bar<T>(std::marker::PhantomData<T>);
+
+impl<T> Trait for Bar<T>
+where
+    T: TypeInfo + 'static,
+    // Foo<T>: TypeInfo + 'static, // adding this bound will fix the overflow error:
```
dvdplm, line 24: Without this bound we should get a compilation error, instead we get infinite recursion.

dvdplm (commented):
I made a smaller repro:

```rust
#[test]
fn recursion_repro4() {
    trait Trait { type Type: TypeInfo + 'static; }

    // Note: not deriving `Encode` for Foo or Bar.
    #[derive(TypeInfo)]
    // Note: the attribute is still here even though I'm not deriving Encode;
    // without it there's no recursion problem.
    struct Foo<S> { #[codec(compact)] a: S }
    #[derive(TypeInfo)]
    struct Bar<T>(std::marker::PhantomData<T>);
    impl<U> Trait for Bar<U>
    // Note: without this, we get infinite recursion.
    // Note2: it makes sense that it is required: `Trait`'s associated type
    // `Type` requires it. But leaving it out should be a compiler error, not
    // infinite recursion.
    where Foo<U>: TypeInfo + 'static
    {
        type Type = Foo<U>;
    }
}
```
It illustrates that the recursion error, `_::_parity_scale_codec::Compact<_>: Decode`, does not stem from deriving `Encode` or `Decode`. Maybe it's the implementation of [`impl<T> Decode for Compact<T>`](https://github.com/paritytech/parity-scale-codec/blob/5593ee726f3f7b03ef20e1bb9e3ae8b9ef029b9e/src/compact.rs#L158) that is the problem here?

thiolliere:
the most basic example I can see is this:
```rust
fn foo<T: scale::HasCompact>() {}
fn bar<T>() { foo::<T>() }
```

I actually don't exactly know why the compiler failing here.

thiolliere:
if I remove CompactAs trait in parity-scale-codec there is no longer recursion error on decode:
```
error[E0277]: the trait bound `Compact<T>: Decode` is not satisfied
 --> tests/foo.rs:2:15
  |
1 | fn foo<T: parity_scale_codec::HasCompact>() {}
  |           ------------------------------ required by this bound in `foo`
2 | fn bar<T>() { foo::<T>() }
  |               ^^^^^^^^ the trait `Decode` is not implemented for `Compact<T>`
  |
  = note: required because of the requirements on the impl of `HasCompact` for `T`

error[E0277]: the trait bound `T: From<Compact<T>>` is not satisfied
 --> tests/foo.rs:2:21
  |
1 | fn foo<T: parity_scale_codec::HasCompact>() {}
  |           ------------------------------ required by this bound in `foo`
2 | fn bar<T>() { foo::<T>() }
  |                     ^ the trait `From<Compact<T>>` is not implemented for `T`
  |
  = note: required because of the requirements on the impl of `Into<T>` for `Compact<T>`
  = note: required because of the requirements on the impl of `HasCompact` for `T`
help: consider restricting type parameter `T`
  |
2 | fn bar<T: From<Compact<T>>>() { foo::<T>() }
  |         ^^^^^^^^^^^^^^^^^^

error[E0277]: the trait bound `parity_scale_codec::compact::CompactRef<'_, T>: WrapperTypeEncode` is not satisfied
 --> tests/foo.rs:2:15
  |
1 | fn foo<T: parity_scale_codec::HasCompact>() {}
  |           ------------------------------ required by this bound in `foo`
2 | fn bar<T>() { foo::<T>() }
  |               ^^^^^^^^ the trait `WrapperTypeEncode` is not implemented for `parity_scale_codec::compact::CompactRef<'_, T>`
  |
  = note: required because of the requirements on the impl of `Encode` for `parity_scale_codec::compact::CompactRef<'_, T>`
  = note: required because of the requirements on the impl of `for<'a> EncodeAsRef<'a, T>` for `Compact<T>`
  = note: required because of the requirements on the impl of `HasCompact` for `T`

error: aborting due to 3 previous errors
```

and if I keep CompactAs trait but I remove the wrappertypedecode implementation:
```rust
impl<T, X> Decode for X where
	T: Decode + Into<X>,
	X: WrapperTypeDecode<Wrapped=T>,
{
	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
		input.descend_ref()?;
		let result = Ok(T::decode(input)?.into());
		input.ascend_ref();
		result
	}
}
```
Then there is recursion error but with a better error message:
```
error[E0275]: overflow evaluating the requirement `Compact<_>: Decode`
 --> tests/foo.rs:2:15
  |
1 | fn foo<T: parity_scale_codec::HasCompact>() {}
  |           ------------------------------ required by this bound in `foo`
2 | fn bar<T>() { foo::<T>() }
  |               ^^^^^^^^
  |
  = help: consider adding a `#![recursion_limit="256"]` attribute to your crate (`foo`)
  = note: required because of the requirements on the impl of `Decode` for `Compact<_>`
  = note: 125 redundant requirements hidden
  = note: required because of the requirements on the impl of `Decode` for `Compact<T>`
  = note: required because of the requirements on the impl of `HasCompact` for `T`
```

thiolliere:
I opened an issue on rustc, for what it worth https://github.com/rust-lang/rust/issues/81785

ascjones:
I've done a new release which includes a "known issue" in the README about this. Will keep an eye in the rust issue.

PR #71, opened 2021-03-02
Remove type parameter defaults

dvdplm:
Defaults for type parameter bounds are only allowed in structs, enums, type and trait definitions, not in impl blocks, so when deriving the `TypeInfo` impl we have to remove them.

See https://github.com/rust-lang/rust/issues/36887 for details.﻿

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![cfg_attr(not(feature = "std"), no_std)]
+// #![cfg_attr(not(feature = "std"), no_std)]
 
 extern crate alloc;
 extern crate proc_macro;
@@ -74,18 +74,14 @@ fn generate(input: TokenStream2) -> Result<TokenStream2> {
 }
 
 fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
-    let mut ast: DeriveInput = syn::parse2(input.clone())?;
+    let ast: DeriveInput = syn::parse2(input.clone())?;
 
     let scale_info = crate_name_ident("scale-info")?;
     let parity_scale_codec = crate_name_ident("parity-scale-codec")?;
 
     let ident = &ast.ident;
 
-    ast.generics
-        .lifetimes_mut()
-        .for_each(|l| *l = parse_quote!('static));
-
-    let (_, ty_generics, _) = ast.generics.split_for_impl();
+    let (impl_generics, ty_generics, _) = ast.generics.split_for_impl();
     let where_clause = trait_bounds::make_where_clause(
         ident,
         &ast.generics,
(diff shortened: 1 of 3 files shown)


derive/src/lib.rs
```diff
@@ -101,12 +101,17 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
...
     let build_type = match &ast.data {
         Data::Struct(ref s) => generate_composite_type(s, &scale_info),
         Data::Enum(ref e) => generate_variant_type(e, &scale_info),
         Data::Union(_) => return Err(Error::new_spanned(input, "Unions not supported")),
     };
+
+    // Remove any type parameter defaults, we don't want those. E.g. `impl<T: Stuff = WutEven>`
+    ast.generics.type_params_mut().for_each(|type_param| {
+        type_param.default = None;
+    });
```
ascjones, line 114: Is there a way to use `ImplGenerics` from `split_for_impl()` above to avoid  this? Not sure whether it will work with our modified lifetimes or not.

dvdplm: You are correct that `ImplGenerics` takes care of [removing the defaults](https://docs.rs/syn/1.0.61/src/syn/generics.rs.html#1068) and it takes care of const parameter defaults too, so it's strictly better.

However there's another problem: we replace all lifetimes with `'static` and this means that given this input:

```
struct Goat<'bar, T: SomeTrait = SomeDefault>
```

 we get this in the expansion:

```
impl<'static, T: SomeTrait>
```

So `ImplGenerics` is the right tool for the job, but using it will require cloning&munging elsewhere. Not sure what is least worse tbh. :/

ascjones: I have an idea: remove the replacing of lifetimes and add something like this to `make_where_clause`:

```
for lifetime in generics.lifetimes() {
        where_clause
            .predicates
            .push(parse_quote!(#lifetime : 'static));
    }
```
That way we should be able to use `ImplGenerics` and remove completely the two mutable bits.

dvdplm: You are completely right. And as an extra bonus we now don't need to clone the input anymore. :heart:

dvdplm: With your suggestion we get an expansion like this:

```rust
    struct Assoc<'bar, T: Types> {
        a: T::A,
        b: &'bar u64,
    }
    
    const _: () = {
        impl<'bar, T: Types> ::scale_info::TypeInfo for Assoc<'bar, T>
        where
            'bar: 'static,
            T::A: ::scale_info::TypeInfo + 'static,
            T: Types + ::scale_info::TypeInfo + 'static,
			…
			…
```

That `'bar: 'static` looks a bit funny but it's valid syntax so all good.

derive/src/lib.rs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![cfg_attr(not(feature = "std"), no_std)]
+// #![cfg_attr(not(feature = "std"), no_std)]
```
ascjones, line 15: ?

ascjones (approved):
LGTM

derive/src/lib.rs
dvdplm: Ouch, my bad.

But while we're at it: we should probably remove this anyway. The derive crate doesn't need to be no-std I think (only its output must be no-std). And as we're using `proc-macro-crate` already it's perhaps good to acknowledge the fact that `scale-info-derive` isn't actually no-std?

ascjones: Yeah I'm not sure why that's there in the first place, but if we remove it should not be hidden in this PR

dvdplm: See https://github.com/paritytech/scale-info/pull/69#discussion_r587434349 – I removed it in that PR.

PR #72, opened 2021-03-03
Derive TypeInfo for own types

dvdplm:
Builds upon/supersedes #21

Derive `TypeInfo` for all of `scale-info`s own types.
﻿

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,9 @@ std = [
 derive = [
     "scale-info-derive"
 ]
-# enables decoding and deserialization of portable scale-info type metadata
+# Derive `TypeInfo` for our own types.
+dogfood = ["derive"]
+# Enables decoding and deserialization of portable scale-info type metadata.
 decode = []
 
 [workspace]
@@ -36,3 +38,4 @@ members = [
     "derive",
     "test_suite",
 ]
+# exclude = ["test_suite"]
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![cfg_attr(not(feature = "std"), no_std)]
+// #![cfg_attr(not(feature = "std"), no_std)]
 
 extern crate alloc;
 extern crate proc_macro;
@@ -76,7 +76,29 @@ fn generate(input: TokenStream2) -> Result<TokenStream2> {
 fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
     let mut ast: DeriveInput = syn::parse2(input.clone())?;
 
-    let scale_info = crate_name_ident("scale-info")?;
+    // Find the name given to the `scale-info` crate in the context we run in.
+    // If scale-info is not among the dependencies then we must be deriving
+    // types for the scale-info crate itself, in which case we need to rename
+    // "self" to something, so the object paths keep working.
+    let (scale_info, import_self_as_scale_info) = {
+        let actual_crate_name = proc_macro_crate::crate_name("scale-info");
+        if let Err(e) = actual_crate_name {
+            if e.starts_with("Could not find `scale-info`") {
+                (Ident::new("_scale_info", Span::call_site()), true)
+            } else {
+                return Err(syn::Error::new(Span::call_site(), e))
+            }
+        } else {
+            (
+                Ident::new(
+                    &actual_crate_name.expect("Checked Err above; qed"),
+                    Span::call_site(),
+                ),
+                false,
+            )
+        }
+    };
+
     let parity_scale_codec = crate_name_ident("parity-scale-codec")?;
 
     let ident = &ast.ident;
@@ -97,33 +119,44 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
     let generic_type_ids = ast.generics.type_params().map(|ty| {
         let ty_ident = &ty.ident;
         quote! {
-            :: #scale_info ::meta_type::<#ty_ident>()
+            #scale_info::meta_type::<#ty_ident>()
         }
     });
 
-    let ast: DeriveInput = syn::parse2(input.clone())?;
+    let mut ast: DeriveInput = syn::parse2(input.clone())?;
     let build_type = match &ast.data {
         Data::Struct(ref s) => generate_composite_type(s, &scale_info),
         Data::Enum(ref e) => generate_variant_type(e, &scale_info),
         Data::Union(_) => return Err(Error::new_spanned(input, "Unions not supported")),
     };
+
+    // Remove any type parameter defaults, we don't want those. E.g. `impl<T: Stuff = WutEven>`
+    ast.generics.type_params_mut().for_each(|type_param| {
+        type_param.default = None;
+    });
     let generic_types = ast.generics.type_params();
     let type_info_impl = quote! {
-        impl <#( #generic_types ),*> :: #scale_info ::TypeInfo for #ident #ty_generics #where_clause {
+        impl <#( #generic_types ),*> #scale_info::TypeInfo for #ident #ty_generics #where_clause {
             type Identity = Self;
-            fn type_info() -> :: #scale_info ::Type {
-                :: #scale_info ::Type::builder()
-                    .path(:: #scale_info ::Path::new(stringify!(#ident), module_path!()))
-                    .type_params(:: #scale_info ::prelude::vec![ #( #generic_type_ids ),* ])
+            fn type_info() -> #scale_info::Type {
+                #scale_info::Type::builder()
+                    .path(#scale_info::Path::new(stringify!(#ident), module_path!()))
+                    .type_params(#scale_info::prelude::vec![ #( #generic_type_ids ),* ])
                     .#build_type
             }
-        }
+        };
     };
-
+    let crate_rename = if import_self_as_scale_info {
+        quote! { extern crate self as _scale_info; }
+    } else {
+        quote! {}
+    };
     Ok(quote! {
-        #[allow(non_upper_case_globals, unused_attributes, unused_qualifications)]
+        // TODO: are these needed?
+        // #[allow(non_upper_case_globals, unused_attributes, unused_qualifications)]
         const _: () = {
-            #type_info_impl;
+            #crate_rename
+            #type_info_impl
         };
     })
 }
@@ -222,7 +255,7 @@ fn generate_composite_type(data_struct: &DataStruct, scale_info: &Ident) -> Toke
         }
     };
     quote! {
-        composite(:: #scale_info ::build::Fields::#fields)
+        composite(#scale_info::build::Fields::#fields)
     }
 }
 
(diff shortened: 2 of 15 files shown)


derive/src/lib.rs
```diff
@@ -97,33 +119,44 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
     let generic_type_ids = ast.generics.type_params().map(|ty| {
         let ty_ident = &ty.ident;
         quote! {
-            :: #scale_info ::meta_type::<#ty_ident>()
+            #scale_info::meta_type::<#ty_ident>()
```
Robbepop, line 122: as before for all non-`scale-info` crates this should still prepend `::` to signal root crate path.

derive/src/lib.rs
```diff
@@ -76,7 +76,29 @@ fn generate(input: TokenStream2) -> Result<TokenStream2> {
...
-    let scale_info = crate_name_ident("scale-info")?;
+    // Find the name given to the `scale-info` crate in the context we run in.
+    // If scale-info is not among the dependencies then we must be deriving
+    // types for the scale-info crate itself, in which case we need to rename
+    // "self" to something, so the object paths keep working.
+    let (scale_info, import_self_as_scale_info) = {
+        let actual_crate_name = proc_macro_crate::crate_name("scale-info");
+        if let Err(e) = actual_crate_name {
+            if e.starts_with("Could not find `scale-info`") {
+                (Ident::new("_scale_info", Span::call_site()), true)
```
Robbepop, line 87: Why use `_scale_info` instead of just `scale_info`?

derive/src/lib.rs
```diff
@@ -76,7 +76,29 @@ fn generate(input: TokenStream2) -> Result<TokenStream2> {
...
+            } else {
+                return Err(syn::Error::new(Span::call_site(), e))
+            }
+        } else {
+            (
+                Ident::new(
+                    &actual_crate_name.expect("Checked Err above; qed"),
```
Robbepop, line 94: Could use better error message. I find `unwrap_or_else(|p| panic!(..))` to deliver far better error messages than `expect`.

derive/src/lib.rs
dvdplm: It doesn't compile with the `::` and I don't know why tbh, but I think that when deriving for `scale-info` itself, and we have `extern crate self as _scale_info;`, then `::_scale_info::path::to::something` doesn't work. :/

derive/src/lib.rs
dvdplm: No good reason, changing.

derive/src/lib.rs
dvdplm: Don't you think that's overkill? This could be `unwrap()` – the error case is checked a few lines above so the comment is for the code reader trained to balk at `unwrap()`s.

derive/src/lib.rs
Robbepop: The `extern crate self as _scale_info;` is not required since Rust edition 2018 and I don't think we should support Rust edition 2015 tbh because it actually resolves many problems we are facing otherwise such as this one.

Robbepop: So either this crate should produce: `crate::` paths or whatever crate alias a dependency uses for `scale-info` prepended with `::`.

Robbepop: To be more concrete: While the `proc-macro-crate` crate is actually useful in these cases it makes use of file I/O while expanding a proc. macro. There are discussions about encapsulating proc. macros in stricter environments that would break proc. macros such as these. Generally tooling such as `rust-analyzer` has good reasons why proc. macros that are "impure" should not be relied on.

dvdplm: > To be honest I am not sure how counterintuitive it is to use the derive macros for the crate that exposes them.

How do you mean "counterintuitive"? I take it you mean it's forcing things a bit?

I think it's a good litmus test for the crate, especially since the type zoo in `scale-info` is fairly advanced and gives us a "free test suite" of sorts. The amount of manual implementations we'd have to maintain is pretty large so if we can derive most of them that's a win imo.

I am concerned by the increased complexity this brings. I will see what I can do to make things a wee bit less messy.

dvdplm: > I am concerned by the increased complexity this brings. I will see what I can do to make things a wee bit less messy.

I think [this commit](https://github.com/paritytech/scale-info/pull/72/commits/b1d26c27db55b638dbfbfaa6f4ef943fff04829a) is makes things much cleaner.

Robbepop: Proc. macros bring overhead with them. So some people prefer using crates but not their provided proc. macros which is why we find `derive` crate feature on some crates to opt-out. It is just nice to be able to separate concerns between the crate that provides definitions and the crate that provides the proc. macro implementations. As soon as the crate providing definitions makes use of the proc. macro you no longer have the separation and a direct dependency instead. I personally am not against this but overhead can be serious in some cases.
The biggest problem is that derive macros that can be used today in their parent crate require this type of "hack" that is encapsulated by the `proc-macro-crate` crate currently. Using this "hack" in any proc. macro turns the proc. macro impure. For example it can no longer enjoy acceleration by the techniques behind the `watt` crate which some people believe to be the future compilation model for all proc. macros since it is super fast and more secure than today's proc. macros.

dvdplm:
@Robbepop @ascjones Do you have any thought about what to do about https://github.com/paritytech/scale-info/blob/5a165b18818c7b15935fc702fbe2167f134a1fbc/src/meta_type.rs#L54-L67 ?

derive/src/lib.rs
dvdplm: 1. The use of `proc-macro-crate` predates this PR so the technical debt is already there;
2. This PR does not introduce a mandatory dependence on `scale-info-derive`: it's opt-in through a feature (as it should be, couldn't agree with you more);

I think we should have a conversation about the use of `proc-macro-crate`; you make a really good point about it making the crate "impure" and we should hash out the trade-offs involved. But I'm not sure this PR is the right place for that?

Robbepop:
> @Robbepop @ascjones Do you have any thoughtsabout what to do about
> 
> https://github.com/paritytech/scale-info/blob/5a165b18818c7b15935fc702fbe2167f134a1fbc/src/meta_type.rs#L54-L67
> ?

Hmmm it is a tricky one. Do we actually need to derive for it? It is only really used as generic parameter in a raw function signature type so maybe `PhantomData` kind is enough?

ascjones:
`MetaType` is never SCALE encoded so shouldn't need `TypeInfo` at all. We only need the `PortableForm`

test_suite/Cargo.toml
```diff
@@ -11,6 +11,8 @@ license = "Apache-2.0"
 
 [dependencies]
 scale-info = { path = "..", features = ["derive", "serde"] }
+# TODO: this should work
```
dvdplm, line 14: I believe @ascjones has a fix in the works for this.

ascjones: Yes I'm working on it

ascjones: https://github.com/paritytech/scale-info/pull/74

Robbepop:
This seems to be stable. Is there a good reason to keep this open and eventually merge it or should we not pursue it further?

dvdplm:
> This seems to be stable. Is there a good reason to keep this open and eventually merge it or should we not pursue it further?

I think we should merge it. It's not super useful in itself, but does provide a sanity check. Happy to update it; won't make a fuss if @ascjones feel it's not needed any longer. :)

ascjones:
It's absolutely something we will need, it just wasn't on the critical path for substrate integration so I have neglected looking at it.

Happy to give this a review once it is updated with `master` and we can get it in in time for the next release.

Robbepop:
We'd need to make extra sure that before/after semantics are equal with respect to the newly derived traits.

Cargo.toml
```diff
@@ -30,6 +30,11 @@ std = [
 derive = [
     "scale-info-derive"
 ]
+# Derive `TypeInfo` for our own types.
+dogfood = [
+    "derive"
+]
+# Enables decoding and deserialization of portable scale-info type metadata.
```
ascjones, line 37: Does this comment belong on top of the feature?

derive/src/lib.rs
```diff
@@ -140,6 +138,25 @@ fn crate_name_ident(name: &str) -> Result<Ident> {
         .map_err(|e| syn::Error::new(Span::call_site(), &e))
 }
 
+/// Find the name given to the `scale-info` crate in the context we run in.
+/// If scale-info is not among the dependencies then we must be deriving
+/// types for the scale-info crate itself, in which case we need to rename
+/// "self" to something, so the object paths keep working.
```
ascjones, line 144: This comment isn't accurate.

derive/src/lib.rs
```diff
@@ -140,6 +138,25 @@ fn crate_name_ident(name: &str) -> Result<Ident> {
...
+fn scale_info_crate_tokens() -> Result<TokenStream2> {
```
ascjones, line 145: We could return `syn::Path` via `parse_quote!` here instead of a `TokenStream`, making it a little more strongly typed, but not necessary.

Cargo.toml
dvdplm: No, I know where this comes from, it's the comment for the "full" feature, with the leading `E` capitalized. Removing.

derive/src/lib.rs
dvdplm: Good catch. Fixing.

ascjones:
Use case for this would be generating equivalent types in another language e.g. https://github.com/polkadot-js/api/blob/58cc197c890881db5a604a4ad98c2446a16e8631/packages/types/src/interfaces/scaleInfo/types.ts.

derive/src/lib.rs
Robbepop: Then we could also rename the function to `scale_info_crate_path` which is also more appealing and reusable.

.github/workflows/rust.yml
```diff
@@ -40,6 +40,7 @@ jobs:
     - name: check-features
       run: |
         cargo check --no-default-features --features bit-vec
+        cargo check --no-default-features --features dogfood
```
Robbepop, line 43: Do we want to keep this optional crate feature? What purpose does it have?

dvdplm: That is a good question. I fail to see the reason to keep this as a feature. @ascjones you ok making it non-optional?

derive/src/lib.rs
```diff
@@ -294,6 +302,23 @@ impl TypeInfoImpl {
...
+//     let crate_ident = match proc_macro_crate::crate_name(SCALE_INFO_CRATE_NAME) {
+//         Ok(FoundCrate::Itself) => parse_quote! { crate },
+//         Ok(FoundCrate::Name(name)) => {
+//             let crate_ident = Ident::new(&name, Span::call_site());
+//             parse_quote! { :: #crate_ident }
+//         }
+//         Err(e) => return Err(syn::Error::new(Span::call_site(), e)),
+//     };
+//     Ok(crate_ident)
+
```
gilescope, line 321: Wasn't entirely sure which version of this function to use so went with master. If that's fine we can delete this commented out version.

gilescope:
Have just brought this up to date as realised I might need this to decode metadata in the browser in rust.

PR #76, opened 2021-03-09
Fix trait bounds for associated types

### LINKED ISSUE #73: Derive does not add bounds for wrapped associated types correctly
The following struct contains an associated type that needs `TypeInfo + 'static` bounds. The derived code does not provide the required bound.

```rust
    trait Colorizer {
        type Property;
    }
    struct Felt {}
    impl Colorizer for Felt {
        type Property = String;
    }
    #[allow(unused)]
    #[derive(TypeInfo)]
    struct Pen<P: Colorizer = Felt> {
        usage: Docs<P>,
        properties: Vec<P::Property>
    }
``` 

`Pen` needs the following bounds:

```
Docs<P>: TypeInfo + 'static,
P: Colorizer + TypeInfo + 'static,
P::Property: TypeInfo + 'static,
```

The derive macro however does not provide the last one, `P::Property: TypeInfo + 'static`.

Macro expansion:

```rust
    const _: () = {
        impl<P: Colorizer> ::scale_info::TypeInfo for Pen<P>
        where
            Docs<P>: ::scale_info::TypeInfo + 'static,
            Vec<P::Property>: ::scale_info::TypeInfo + 'static,
            P: Colorizer + ::scale_info::TypeInfo + 'static,
        {
            type Identity = Self;
                ::scale_info::Type::builder()
                    .path(::scale_info::Path::new("Pen", "derive"))
                    .type_params(<[_]>::into_vec(box [::scale_info::meta_type::<P>()]))
                    .composite(
                        ::scale_info::build::Fields::named()
                            .field_of::<Docs<P>>("usage", "Docs<P>")
                            .field_of::<Vec<P::Property>>("properties", "Vec<P::Property>"),
                    )
            }
        };
    };
```

A concrete example of where we need this is https://github.com/paritytech/scale-info/blob/master/src/ty/mod.rs#L65-L81

Notice the odd `Vec<P::Property>: ::scale_info::TypeInfo + 'static,` in the expansion.

dvdplm:
When a type has a type parameter that is bound to a trait with an associated type and that associated type has the same name as the type we're deriving TypeInfo for, the trait bounds are wrong.

See #73 for more gory details.

Thanks to @thiolliere for [the fix](https://github.com/paritytech/parity-scale-codec/issues/257).

Closes #73

### DIFF
--- a/derive/src/trait_bounds.rs
+++ b/derive/src/trait_bounds.rs
@@ -18,10 +18,11 @@ use syn::{
     parse_quote,
     punctuated::Punctuated,
     spanned::Spanned,
-    visit::Visit,
+    visit::{self, Visit},
     Generics,
     Result,
     Type,
+    TypePath,
     WhereClause,
 };
 
@@ -112,6 +113,34 @@ fn type_contains_idents(ty: &Type, idents: &[Ident]) -> bool {
     visitor.result
 }
 
+/// Checks if the given type or any containing type path starts with the given ident.
+fn type_or_sub_type_path_starts_with_ident(ty: &Type, ident: &Ident) -> bool {
+    // Visits the ast and checks if the a type path starts with the given ident.
+    struct TypePathStartsWithIdent<'a> {
+        result: bool,
+        ident: &'a Ident
+    }
+
+    impl<'a, 'ast> Visit<'ast> for TypePathStartsWithIdent<'a> {
+        fn visit_type_path(&mut self, i: &'ast TypePath) {
+            if i.qself.is_none() {
+                if let Some(segment) = i.path.segments.first() {
+                    if &segment.ident == self.ident {
+                        self.result = true;
+                        return;
+                    }
+                }
+            }
+            visit::visit_type_path(self, i);
+        }
+    }
+
+	let mut visitor = TypePathStartsWithIdent { result: false, ident };
+	visitor.visit_type(ty);
+	visitor.result
+}
+
+
 /// Returns all types that must be added to the where clause with a boolean
 /// indicating if the field is [`scale::Compact`] or not.
 fn collect_types_to_bind(
@@ -128,7 +157,7 @@ fn collect_types_to_bind(
                 &&
                 // Remove all remaining types that start/contain the input ident
                 // to not have them in the where clause.
-                !type_contains_idents(&field.ty, &[input_ident.clone()])
+                !type_or_sub_type_path_starts_with_ident(&field.ty, &input_ident)
             })
             .map(|f| (f.ty.clone(), super::is_compact(f)))
             .collect()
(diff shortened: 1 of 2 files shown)


thiolliere (approved):

Robbepop:
I won't stop you from merging this but will definitely tell you (again) that fixes like these are not implementation details in a proc. macro but will from now on be part of its API. So people are going to depend on these semantics that are based on syntactic analysis of the code. This is always pretty fragile and it is not really possible to catch all the cutting edges. What about type alises? What about nested associated types? What about macro types? Etc.

So this is at best a best-effort approach and not an ideal solution. As it I would rather see more momentum in making it possible for users to define custom trait bounds. This would equally solve the same problems (and more since it is a more general solution) at the obvious cost of demanding users to provide the trait bounds instead of inferring them for some common cases ... what happens on the other cases though? :thinking:

thiolliere:
This special case is just a best effort to avoid having wrong bounds in recursive types.

But you are right, and now I realize `struct Foo ( Option<Box<Self>> );` or other alias of `Foo` is probably failing too.

This special shouldn't harm too much because any recursive types (which is what we try to detect here, with lot of misses in case of alias, but now there shouldn't be false positive anymore) will have wrong bounds if we derive them normally.

But I actually agree best is to allow user to write bounds. Unfortunately the error message for failing default bounds in recursive types will not be very good, but maybe more understandable than potentially failing best effort.

dvdplm:
> So this is at best a best-effort approach and not an ideal solution.

This is crystal clear to me and this bug here is the perfect illustration of your point (and also, imo, validates the dogfood-work @ascjones started and I am pushing over the finish line).

I think the short/mid-term end-goal of `scale-info` remains the same though: be "good enough" for substrate and polkadot while we think about what a better approach could be.

> As it I would rather see more momentum in making it possible for users to define custom trait bounds.

Adding support for custom bounds is the highest priority. :)
If we see that approach work out well maybe we can scale back on the syntax hacks but I suspect that asking regular users to type out the bounds manually is going to be a high barrier to adoption.

Robbepop:
> Adding support for custom bounds is the highest priority. :)

So maybe we should wait for this work before we merge this PR because merging this PR and later replacing it with custom trait bounds is both a breaking change due to both PRs being API changes to the proc. macro.

> [..] but I suspect that asking regular users to type out the bounds manually is going to be a high barrier to adoption.

I am more optimistic since it was not an adoption barrier for `serde` as well.

ascjones (approved):
LGTM

Robbepop:
In a meeting it was decided that we put forward with `scale-info` in the following way:

- We will go with the "hacks" that have already been implemented for it and for `parity-scale-codec` crates to cover the most common use cases with 80/20 rule.
- For the other 20% of cases that are not covered by the `scale-info` and `parity-scale-codec` derive macros we will provide a feature for users to define custom trait bounds similar to the design `serde` chose. Those custom trait bounds will be replacing the generated ones instead of being additive.
- We might consider in the future to remove the "cruft" of "hacks" for common cases and consider removing them altogether with a major release for both derives.

PR #77, opened 2021-03-11
Upgrade proc-macro-crate to v1

dvdplm:


derive/src/lib.rs
```diff
@@ -125,7 +125,13 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
 /// Get the name of a crate, to be robust against renamed dependencies.
 fn crate_name_ident(name: &str) -> Result<Ident> {
     proc_macro_crate::crate_name(name)
-        .map(|crate_name| Ident::new(&crate_name, Span::call_site()))
+        .map(|crate_name| {
+            use proc_macro_crate::FoundCrate::*;
+            match crate_name {
+                Itself => Ident::new("self", Span::call_site()),
```
Robbepop, line 131:
```suggestion
                Itself => Ident::new("crate", Span::call_site()),
```

Robbepop: Isn't `crate` the right keyword then? Actually I wonder if `crate` as `Ident` is allowed.

dvdplm: No, it's not, which is why I switched to [a token stream in the other PR](https://github.com/paritytech/scale-info/pull/72/files#diff-60da9bde79deaa50c0ce6f42d1f47205ba31f448dc2713cc746339343581b699R134).

My plan was to merge this change here and then merge it into the other and sort out the `Itself`-case there.

Robbepop (approved):
With https://github.com/paritytech/scale-info/pull/77#discussion_r592703530 said it is an approval from me.

ascjones (approved):

PR #80, opened 2021-03-23
Handle `#[codec(index = …)]` in regular enums

### LINKED ISSUE #79: Handle `#[codec(index = $int)]` on regular enums
PR #44 handles the `index` attribute on C-like enums, but does not address regular enums with some variants marked `#[codec(index = $int)]`.

All enum variants have a discriminant. C-style enums allow explicitly setting the discriminant. Regular enums do not. 
The `#[codec(index = 123)]` attribute sets the index byte for the variant it decorates when the enum is SCALE encoded.

"index" and "discriminant" do not mean the same thing, but the current implementation conflates them somewhat and the default "index" for a variant is, in order of priority, the `#[codec(index = …)]` attribute, the explicitly set discriminant, and lastly the implicit index from the order of appearance.

I believe `parity-scale-codec` is slightly broken today, in that the following will lead to bad things happening when encoding/decoding:

```
#[derive(Encode)]
enum Foo {
	A,
	#[codec(index = 2)]
	B = 8,
	C
}
```

The `Variant` struct in `scale-info` has a `discriminant` member, an `Option<u64>`. Today it is `None` for all non C-style enums.

Picking the right solution here can be a bit tricky. We can use the current `discriminant` field to store the index. This is a questionable solution though, as we perpetuate the mixing of index and discriminant. Open to alternative ideas.

dvdplm:
A Variant can have fields of 3 different kinds: `Named`, `Unnamed` or `Unit`. In this PR I introduce indexed variations of these, so we have `VariantsBuilder::variant`/`VariantsBuilder::indexed_variant` and `VariantsBuilder::variant_unit`/`VariantsBuilder::indexed_variant_unit`.

This PR stores the index in a new [`index`](https://github.com/paritytech/scale-info/blob/33f16f12cc88f3ac82995c6839fa47222ea5573b/src/ty/variant.rs#L166) member of `Variant`, which is probably a bit contentious given it increases the size of the registry. 
We could also store the index in the `discriminant` field, which would be compatible with the direction the Rust language is heading with [RFC 2363](https://github.com/rust-lang/rfcs/pull/2363) (surfacing discriminants for all enum variants, c-style or not).

It is currently possible to set **both** the discriminant and the index on a c-style enum (with priority given to the index attribute):

```
#[derive(Encode)]
enum Foo {
	A,
	#[codec(index = 2)]
	B = 8,
	C
}
```

I think that the more cautious path is to have both `index` and `discriminant`, mimicing `parity-scale-info`s behaviour.

Builds on top of #44 

Closes #79


derive/src/lib.rs
```diff
@@ -245,9 +245,21 @@ fn generate_variant_type(data_enum: &DataEnum, scale_info: &Ident) -> TokenStrea
...
+                    quote! {
+                        .indexed_variant(
+                            #v_name,
+                            #index,
+                            :: #scale_info::build::Fields::named()
+                                #( #fields)*
+                        )
+                    }
+                }
+                (Fields::Named(ref fs), None) => {
```
ascjones, line 262: This whole block would be simplified if `.variant(` accepted an `index: Option<u8>`, instead of having two methods. What do you think?

test_suite/tests/ui/fail_with_invalid_codec_attrs.stderr
```diff
@@ -25,10 +19,10 @@ error: proc-macro derive panicked
    = help: message: scale-info: Bad index in `#[codec(index = …)]`, see `parity-scale-codec` error: Error("expected literal")
 
 error: expected literal
-  --> $DIR/fail_with_invalid_codec_attrs.rs:24:25
+  --> $DIR/fail_with_invalid_codec_attrs.rs:18:21
```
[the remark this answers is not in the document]
dvdplm: Indeed. It is a matter or order: recent rusts put the errors in the order of appearance in the file, so errors on line 18 come before errors on line 24.

Given that the order is more correct on newer rusts, can we bump CI to use a more recent version?

derive/src/lib.rs
dvdplm: I go back and forth on that. This code here sure is ugly; otoh the API itself is straight forward: there's no way users can misunderstand what `indexed_variant()` does.

The best option would be a builder, i.e. to have:

```
Variants::with_fields()
    .variant(
        "A", 
        Fields::unnamed()
            .field_of::<u8>("u8")
    )
    .indexed(123)
```

…but I think that's fiddly to get working.

dvdplm: In other words: if you have a strong opinion, I'll follow your lead! ;)

ascjones: Given that we have another optional field on the way (docs) #87, maybe it is worth now introducing a variant builder. Hopefully that would also simplify this big block of code.

test_suite/tests/ui/fail_with_invalid_codec_attrs.stderr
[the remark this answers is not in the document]
ascjones: I believe the CI automatically uses the latest nightly which includes the required components (`rustfmt`, `clippy`).  I see there was a new one yesterday `1.54.0-nightly (bacf770f2 2021-05-05)`, it might be picked up on a rerun.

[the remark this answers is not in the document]
ascjones: It did pick up the latest one as above, but the UI test still fails, so might need to update those UI tests using `1.54.0-nightly`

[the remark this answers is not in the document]
dvdplm: CI and my rustc seems to be in agreement now. Ty.

derive/src/lib.rs
ascjones: Builder added and merged  from #87 into this branch

derive/src/lib.rs
```diff
@@ -245,39 +260,45 @@ fn generate_variant_type(data_enum: &DataEnum, scale_info: &Ident) -> TokenStrea
         .filter(|v| !utils::should_skip(&v.attrs))
         .map(|v| {
             let ident = &v.ident;
-            let v_name = quote! {stringify!(#ident) };
-            match v.fields {
+            let v_name = quote! {::core::stringify!(#ident) };
+            let docs = utils::get_doc_literals(&v.attrs);
+            let index = utils::maybe_index(v).map(|i| quote!(.index(#i)));
```
dvdplm, line 265: It's hard to overstate how much nicer this is compared to what I had.

ascjones: That was one of my goals with the builder pattern changes, so thanks!, The main priorities being type safe construction and friendliness for the derive macro. The tradeoff being that it is not fantastic for constructing type infos by hand, but we are assuming that is not very common.

ascjones (approved):
LGTM

PR #83, opened 2021-04-12
Remove HasCompact::Type bounds

ascjones:
Removes the requirement for `HasCompact::Type: TypeInfo` bounds I had to add to some impls in substrate, see them removed in this commit https://github.com/paritytech/substrate/commit/7da3bb82351b2f45f12a06b5be64c87f256a9cc8. 

It does this by just registering `T` for a compact field in a `Compact<T>`. This relies on the fact that the [blanket impl](https://github.com/paritytech/parity-scale-codec/blob/master/src/compact.rs#L222) always has `Type = Compact<T>`.  

@dvdplm maybe you already tried this as part of https://github.com/paritytech/scale-info/pull/53 and decided against it.

### DIFF
--- a/derive/src/trait_bounds.rs
+++ b/derive/src/trait_bounds.rs
@@ -71,9 +71,6 @@ pub fn make_where_clause<'a>(
             where_clause
                 .predicates
                 .push(parse_quote!(#ty : :: #parity_scale_codec ::HasCompact));
-            where_clause
-                .predicates
-                .push(parse_quote!(<#ty as :: #parity_scale_codec ::HasCompact>::Type : :: #scale_info ::TypeInfo + 'static));
         } else {
             where_clause
                 .predicates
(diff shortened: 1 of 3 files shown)


dvdplm:
Yeah so this is what I had in my very first implementation. I was wary about it mostly because I had a poor understanding of what `Compact` and `HasCompact` even was; it felt a bit too easy to be correct?

When a type is `#[code(compact)]` it means that instead of encoding the type we actually encode `<the_type as HasCompact>::Type` so it becomes a bit philosophical here: do we want `TypeInfo` for `the_type` or for the type actually used in the encoding, `::Type`? 
At the time of #53 I felt pretty convinced that we wanted `TypeInfo` for the type that would end up being encoded (and then we ran into fun compiler bugs and I didn't think too deeply about it).

If this works for substrate then I'm all for it.

ascjones:
> do we want TypeInfo for the_type or for the type actually used in the encoding, ::Type

`TypeInfo` for `::Type`, however we know this is always `Compact<T>` because of the blanket impl. 

I'll leave this here as a draft for now, and come back to it once the substrate PR is in a good state, and maybe explore other ways to get around having those bounds. And consider whether this is philosophically/practically the right approach.

dvdplm (approved):
LGTM

PR #87, opened 2021-04-26
Capture doc comments, add variant and field builders

ascjones:
The derive macro captures doc comments for types, fields (named and unnamed) and enum variants.

Since docs are optional, builders are introduced for fields and variants. This lays the groundwork for https://github.com/paritytech/scale-info/pull/80 which allows an optional index to be applied to a variant.

Related to https://github.com/paritytech/substrate/pull/8615, This PR will enable removing some custom metadata types in substrate e.g. [`FunctionMetadata`](https://github.com/paritytech/frame-metadata/blob/86bf2b374c9aa7fee82ad19df8b5569b0d1b8c58/frame-metadata/src/v13.rs#L302).

# todo

- [x] fix tests


src/registry.rs
```diff
@@ -265,4 +280,119 @@ mod tests {
             panic!("Should be a composite type definition")
         }
     }
+
+    #[test]
+    fn clear_docs() {
+        #[allow(unused)]
+        /// docs
+        struct S {
+            /// docs
```
Robbepop, line 289: For tests I'd use different doc contents for every documentation line as it is more robust to testing errors.

src/ty/variant.rs
```diff
@@ -222,4 +246,17 @@ where
...
+    pub fn docs(&self) -> &[T::String] {
+        &self.docs
+    }
+
+    /// Clear the docs for this variant.
+    pub fn clear_docs(&mut self) {
+        self.docs.clear();
+        for field in &mut self.fields {
+            field.clear_docs()
+        }
```
Robbepop, line 260: I am not entirely convinced if this strategy of first building up the entire documentation and then getting rid of it again represents an optimal API.

test_suite/tests/derive.rs
```diff
@@ -158,11 +176,11 @@ fn c_like_enum_derive_with_scale_index_set() {
...
-            .variant("A", 0)
-            .variant("B", 10)
-            .variant("C", 13)
-            .variant("D", 3)
-            .variant("E", 14),
+            .variant("A", 0, &[])
+            .variant("B", 10, &[])
+            .variant("C", 13, &[])
+            .variant("D", 3, &[])
+            .variant("E", 14, &[]),
```
Robbepop, line 183: I'd actually prefer an API where I could do:
```rust
.variant("A", 0).docs(&[])
```
instead, where the `docs(..)` is not mandatory. so if you leave it away docs are naturally empty.

src/ty/variant.rs
ascjones: The static alternative I considered was to provide a feature for the derive macro which would not capture the docs in the first place. However if you wanted metadata both with and without docs you would have to compile it twice, so I opted for the dynamic version since metadata serialization should usually be a one time thing.

Can you think of a better way?

src/registry.rs
ascjones: Also this test is actually misleading because the `TypeInfo` is handrolled and not derived.

src/registry.rs
```diff
@@ -189,13 +189,25 @@ impl PortableRegistry {
             (id, ty)
         })
     }
+
+    /// Clears all docs from types.
+    ///
+    /// Use when the docs are not required to produce smaller encoded metadata.
+    pub fn clear_docs(&mut self) {
```
dvdplm, line 196: Not for this PR, but maybe at some point it'd be useful to return only the docs, or a `HashMap<TypeName, Vec<&'static str>>`. For instance in `substrate-api-sidecar` some of our docs are automatically generated from the rustdocs by way of the metadata. Could be useful to get only the docs.

src/registry.rs
```diff
@@ -265,4 +280,119 @@ mod tests {
...
+            pub u: u8,
+        }
+
+        impl TypeInfo for S {
+            type Identity = Self;
+
+            fn type_info() -> Type {
+                Type::builder()
+                    .path(Path::new("S", module_path!()))
+                    .docs(&[" docs"])
```
dvdplm, line 300: This leading space is a bit annoying. We could take an opinionated stance and strip off leading ` ` or `\t`.

test_suite/tests/derive.rs
```diff
@@ -130,14 +140,22 @@ fn unit_struct_derive() {
 fn c_like_enum_derive() {
     #[allow(unused)]
     #[derive(TypeInfo)]
+    /// Enum docs.
     enum E {
+        /// Unit variant.
         A,
+        /// Variant with discriminator.
```
dvdplm, line 147:
```suggestion
        /// Variant with discriminant.
```

test_suite/tests/derive.rs
```diff
@@ -130,14 +140,22 @@ fn unit_struct_derive() {
...
     }
 
     let ty = Type::builder()
         .path(Path::new("E", "derive"))
-        .variant(Variants::fieldless().variant("A", 0u64).variant("B", 10u64));
+        .docs(&[" Enum docs."])
+        .variant(
+            Variants::fieldless()
+                .variant("A", 0u64, &[" Unit variant."])
+                .variant("B", 10u64, &[" Variant with discriminator."]),
```
dvdplm, line 157:
```suggestion
                .variant("B", 10u64, &[" Variant with discriminant."]),
```

test_suite/tests/derive.rs
dvdplm: I agree such an API would be nicer. It would require a fair amount of refactoring to achieve though. For other cases in the past we've said that it's ok-ish to have an ugly API given it is rare for humans to use it: most types are derived.
That said, if it's possible, I'd prefer it as well. We'd likely end up with having to call `finalize()` explicitly (which imo is fine), like `.variant("A", 0).docs(&[]).finalize()` (or something similar).

test_suite/tests/derive.rs
```diff
@@ -172,20 +190,39 @@ fn c_like_enum_derive_with_scale_index_set() {
...
+            T,
+        ),
+        /// Named fields variant.
+        B {
+            /// Named field.
+            b: T,
+        },
+        /// Unit variant.
         C,
     }
```
dvdplm, line 207: Would be good to test for things like:

- `#[doc(hidden)`
- `#[doc = " This is a doc comment."]`
- That we handle inner/outer docs properly, e.g. a misplaced `//! Wut?`
- `#[doc(primitive)]` or any other less common variants of the `#[doc]` attribute?

dvdplm (commented):
This lgtm; the only outstanding issues as far as I can see are:
-  whether we want to refactor the crate to use a more extensive builder-pattern. I had a stab at that a while back and found it tricky.
- a few more tests for the `#[doc]` attribute

derive/src/lib.rs
```diff
@@ -153,16 +153,24 @@ fn generate_fields(fields: &FieldsList) -> Vec<TokenStream2> {
...
+            let type_of_method = if utils::is_compact(f) {
+                quote!(compact)
             } else {
-                quote!(.field_of::<#ty>)
+                quote!(ty)
             };
-            if let Some(ident) = ident {
-                quote!(#method_call(stringify!(#ident), #type_name, &[ #( #docs ),* ]))
+            let name = if let Some(ident) = ident {
+                quote!(.name(stringify!(#ident)))
```
Robbepop, line 162: To make our macros more robust please always use full paths:
https://doc.rust-lang.org/core/macro.stringify.html
E.g. in this case `::core::stringify!(#ident)`

Robbepop: You can apply this to all occurrences of macros, functions etc.

Robbepop: For trait methods we should always use `<X as Y>::z(..)` syntax to disambiguate on a semantic level.

Robbepop: When referring to primitive types such as `bool`, `i32` and `str` we should always expand to types found in https://doc.rust-lang.org/core/primitive/index.html.

ascjones: > For trait methods we should always use `<X as Y>::z(..)` syntax to disambiguate on a semantic level.

Which trait method call are you referring to?

ascjones: > When referring to primitive types such as bool, i32 and str we should always expand to types found in https://doc.rust-lang.org/core/primitive/index.html.

I think this is outside the scope of this PR, it would need some considerations: e.g. how do we determine at the macro level whether a `bool` is indeed a `primitive::bool` and not a shadowing type as per the example.

Robbepop: > Which trait method call are you referring to?

Was just a general rule of thumb.

> I think this is outside the scope of this PR, it would need some considerations: e.g. how do we determine at the macro level whether a bool is indeed a primitive::bool and not a shadowing type as per the example.

No, this is just needed when you literally would expand to a primitive type in a macro.

ascjones: > Was just a general rule of thumb.

Makes sense, but is there an instance here you have spotted where I call a not fully qualified trait method?

> No, this is just needed when you literally would expand to a primitive type in a macro.

I might have misunderstood the comments as being specific to this file/PR, where they are intended rather as further macro good practices.

src/ty/variant.rs
ascjones: I've removed this clearing docs feature, for ease of getting this PR done. We can always add it back or an alternative as a follow up.

ascjones: https://github.com/paritytech/scale-info/issues/94

derive/src/lib.rs
```diff
@@ -96,13 +96,16 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
...
     };
+    let docs = utils::get_doc_literals(&ast.attrs);
+
     let type_info_impl = quote! {
         impl #impl_generics :: #scale_info ::TypeInfo for #ident #ty_generics #where_clause {
             type Identity = Self;
             fn type_info() -> :: #scale_info ::Type {
                 :: #scale_info ::Type::builder()
-                    .path(:: #scale_info ::Path::new(stringify!(#ident), module_path!()))
+                    .path(:: #scale_info ::Path::new(::core::stringify!(#ident), module_path!()))
```
Robbepop, line 106:
```suggestion
                    .path(:: #scale_info ::Path::new(::core::stringify!(#ident), ::core::module_path!()))
```

Robbepop: Also `module_path` and other Rust macros should be used by their full path in macro generated code.

derive/src/lib.rs
```diff
@@ -214,13 +226,18 @@ fn generate_c_like_enum_def(variants: &VariantList, scale_info: &Ident) -> Token
         .map(|(i, v)| {
             let name = &v.ident;
             let discriminant = utils::variant_index(v, i);
+            let docs = utils::get_doc_literals(&v.attrs);
             quote! {
-                .variant(stringify!(#name), #discriminant as u64)
+                .variant(
+                    :: #scale_info ::Variant::builder(::core::stringify!(#name))
+                        .index(#discriminant as u64)
```
Robbepop, line 233:
```suggestion
                        .index(#discriminant as ::core::primitive::u64)
```

Robbepop: this is an instance where you could apply the full path usage from Rust's `core::primitive` module.

ascjones: Ah ha. Thanks!

src/registry.rs
ascjones: Mentioned in https://github.com/paritytech/scale-info/issues/94

src/registry.rs
ascjones: Changed it to remove the leading space, maybe not the best solution but it works. Leaving tabs because they might be for formatting.

src/ty/variant.rs
```diff
@@ -222,4 +225,9 @@ where
     pub fn discriminant(&self) -> Option<u64> {
         self.discriminant
     }
+
+    /// Returns the documentation of the variant.
+    pub fn docs(&self) -> &[T::String] {
+        &self.docs
```
dvdplm, line 231: Maybe `doc` should be an `Option`. Feels a bit more rusty to check for `None` rather than an empty slice?

src/build.rs
```diff
@@ -251,110 +260,234 @@ impl<T> FieldsBuilder<T> {
...
+            FieldBuilder,
+        )
+            -> FieldBuilder<field_state::NameAssigned, field_state::TypeAssigned>,
     {
-        self.fields.push(Field::named_of::<T>(name, type_name));
+        let builder = builder(FieldBuilder::new());
+        self.fields.push(builder.finalize());
         self
     }
+}
```
dvdplm, line 275: The main concern I have is how the api becomes a bit clunky to use with this, all the `Fn(FieldBuilder)` makes things a bit hard to read. I'm not sure if it's a good idea or not but maybe we could have convenience methods like e.g. 

```rust
    pub fn field_of<T: TypeInfo + 'static>(self, name: &'static str, type_name: &'static str) -> Self {
        self.field(|f| f.ty::<T>().name(name).type_name(type_name))
    }
```
…to cover the common cases? Maybe it's a terrible idea.

dvdplm (commented):
I think the `FieldsBuilder<N, T>` state machinery is really clever, good job there. 
The resulting API is a bit awkward to read but I'm not sure how difficult it is to do better. Also had a question about what it would take to always populate the discriminant member for variants.

src/build.rs
```diff
@@ -251,110 +260,234 @@ impl<T> FieldsBuilder<T> {
...
+        self.fields.push(builder.finalize());
         self
     }
 }
 
-impl FieldsBuilder<UnnamedFields> {
-    /// Add an unnamed field with the type of the type parameter `T`
-    pub fn field_of<T>(mut self, type_name: &'static str) -> Self
+/// Type states for building a field.
+pub mod field_state {
```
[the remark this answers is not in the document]
ascjones: Using the type state builder pattern to ensure that only all named or all unnamed fields can be constructed. The other option is a runtime check, which would make for less code but better to catch things at compile time.

src/build.rs
```diff
@@ -251,110 +260,234 @@ impl<T> FieldsBuilder<T> {
...
-impl<T> VariantsBuilder<T> {
-    fn new() -> Self {
-        VariantsBuilder {
-            variants: Vec::new(),
-            marker: Default::default(),
+/// Build a [`Variant`].
+pub struct VariantBuilder {
+    name: &'static str,
+    fields: Vec<Field<MetaForm>>,
+    discriminant: Option<u64>,
```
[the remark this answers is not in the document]
ascjones: Possibly. If we do we should do it as part of #80 though.

src/ty/variant.rs
ascjones: I prefer an empty slice, it conveys the same meaning and is easier to consume.

src/build.rs
ascjones: Could add those for use in tests and handrolled impls, but the primary consumer of this API is the derive macro which would always use the builder pattern.

This PR did originally just add the docs to those helper methods, but has changed to the builder pattern after suggestions in this review. I agree it is slightly more clunky but also more flexible and doesn't require all the permutations of the helper methods, especially when we add optional indices in #80.

dvdplm: Yeah, I struggled with this too. I have no better suggestion atm. :/

Definitely agree on the flexibility. We can add the convenience methods if/when we feel we need them.

dvdplm (approved):

PR #88, opened 2021-05-01
Add new top-level attribute `scale_info(bounds(T: SomeTrait + OtherTrait))`

### LINKED ISSUE #78: Trait bounds: derive as much as is reasonable, `#[scale_info(bounds())]` for the rest
In a meeting it was decided that we put forward with `scale-info` in the following way:

- We will go with the "hacks" that have already been implemented for it and for `parity-scale-codec` crates to cover the most common use cases with 80/20 rule.
- For the other 20% of cases that are not covered by the `scale-info` and `parity-scale-codec` derive macros we will provide a feature for users to define custom trait bounds similar to the design `serde` chose. Those custom trait bounds will be replacing the generated ones instead of being additive.
- We might consider in the future to remove the "cruft" of "hacks" for common cases and consider removing them altogether with a major release for both derives.

_Originally posted by @Robbepop in https://github.com/paritytech/scale-info/pull/76#issuecomment-797506544_

dvdplm:
Inspired by https://github.com/paritytech/parity-scale-codec/pull/263, this PR adds a `#[scale_info(bounds(…))]` attribute.

Closes #78 

Outstanding: come up with relevant test(s).

### DIFF
--- a/test_suite/tests/derive.rs
+++ b/test_suite/tests/derive.rs
@@ -442,6 +442,42 @@ fn whitespace_scrubbing_works() {
     assert_type!(A, ty);
 }
 
+#[test]
+fn custom_bounds() {
+    // TODO: this test is dumb. It's a copy of Basti's equivalent in `parity-scale-codec` but I
+    // don't think it can work for us. I need a proper example of when custom bounds are needed.
+    // As-is, this test is simply setting the same bounds as the derive would have, which is pretty
+    // pointless.
+    #[allow(unused)]
+    #[derive(TypeInfo)]
+    #[scale_info(bounds(T: Default + TypeInfo + 'static, N: TypeInfo + 'static))]
+    struct Hey<T, N> {
+        ciao: Greet<T>,
+        ho: N,
+    }
+
+    #[derive(TypeInfo)]
+    #[scale_info(bounds(T: TypeInfo + 'static))]
+    struct Greet<T> {
+        marker: PhantomData<T>,
+    }
+
+    #[derive(TypeInfo)]
+    #[derive(Default)]
+    struct SomeType;
+
+    let ty = Type::builder()
+        .path(Path::new("Hey", "derive"))
+        .type_params(tuple_meta_type!(SomeType, u16))
+        .composite(
+            Fields::named()
+                .field_of::<Greet<SomeType>>("ciao", "Greet<T>")
+                .field_of::<u16>("ho", "N")
+        );
+
+    assert_type!(Hey<SomeType, u16>, ty);
+}
+
 #[rustversion::nightly]
 #[test]
 fn ui_tests() {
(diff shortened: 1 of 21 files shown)


test_suite/tests/derive.rs
```diff
@@ -442,6 +442,41 @@ fn whitespace_scrubbing_works() {
     assert_type!(A, ty);
 }
 
+#[test]
+fn custom_bounds() {
+    // TODO: this test is dumb. It's a copy of Basti's equivalent in `parity-scale-codec` but I
+    // don't think it can work for us. I need a proper example of when custom bounds are needed.
+    // As-is, this test is simply setting the same bounds as the derive would have, which is pretty
+    // pointless.
```
ascjones, line 450: This is the crux of the issue, what is a valid use case for this feature?

An important difference here between  `scale-info` and `parity-scale-codec` is that it has requirement that any type parameters also implement `TypeInfo` because the derive always adds them to `type_params`. This is required for e.g. a Rust type generator where we need to be able to generate a generic types. 

So in your example below, the `codec` generated derive does not require `Encode`/`Decode` for `T` but would for `N`. So that's an area where you could use the custom bounds to enable/disable the bound on `T`.  Whereas `TypeInfo` is always required for `N` *and* `T` so there's no use for the custom bounds in that case.

ascjones: I agree with the principle of having this [escape hatch](https://serde.rs/attr-bound.html) for corner cases, but with the restriction of type params always needing to be `TypeInfo` the flexibility is limited. 

Perhaps there is a  test case to be found in using it with associated types to override the auto bounds e.g.

```
trait T { type A; }

struct S<T: Trait> {
  a: Inner<T::A>,
}

struct Inner<T>(PhantomData<T>);

impl<T> TypeInfo for Inner<T> {
 ... /// manual impl which doesn't require T: TypeInfo
}
```

dvdplm: > This is the crux of the issue, what is a valid use case for this feature?

Head on the nail. I thought we'd be able to sort out those pesky recursion cases with this, but when I started looking at the repros we have for that I realised this feature here does no good for those. Obvious in hindsight.

> Perhaps there is a test case to be found in using it with associated types to override the auto bounds 

Good idea, I'll try that.

dvdplm: Found [this](https://serde.rs/attr-bound.html#handwritten-generic-type-bounds) in the serde docs. Might be an idea for a good test case too.

ascjones: Here we go: https://github.com/paritytech/scale-info/pull/88/commits/3a987993add90c83d783ee75cfe89d1faca16877

ascjones: I suggest removing this test, and just using a `ui` pass test with the attribute, since comparing the derived `TypeInfo` doesn't really add any value here.

ascjones: Done https://github.com/paritytech/scale-info/pull/88/commits/c35145da9a26affd72e1fcf2bc2b9ad78f044d71

ascjones (approved):

PR #96, opened 2021-05-28
Add `skip_type_params` attribute

ascjones:
Currently the `TypeInfo` derive [adds all generic type params](https://github.com/paritytech/scale-info/blob/master/derive/src/lib.rs#L95) to the [`Type::type_params`](https://github.com/paritytech/scale-info/blob/master/src/ty/mod.rs#L77) metadata.

Therefore all type params must be constrained to `TypeInfo` in order for the call to [`meta_type`](https://github.com/paritytech/scale-info/blob/master/src/lib.rs#L158) to compile. So the proc macro [automatically adds](https://github.com/paritytech/scale-info/blob/master/derive/src/trait_bounds.rs#L82) those bounds.

However, sometimes this requirement forces us to implement `TypeInfo` or provide additional bounds where it is unnecessary. 

This PR introduces the `skip_type_params` attribute, which removes the requirement for the supplied type params to implement `TypeInfo`.

## Scenario 1: `skip_type_params` on its own.

Consider a  simple example of a type parameter which is used for associated types, but the type itself does not carry any type information. This is a common pattern in substrate:

```rust
trait Config {
    type Balance;
}

struct Runtime; // doesn't implement `TypeInfo`

impl Config for Runtime {
    type Balance = u64;
}

#[allow(unused)]
#[derive(TypeInfo)]
#[scale_info(skip_type_params(T))] 
struct A<T: Config> {
    balance: T::Balance,
    marker: core::marker::PhantomData<T>,
}

fn assert_type_info<T: TypeInfo + 'static>() {}

fn main() {
    assert_type_info::<A<Runtime>>(); // without the `skip_type_params` attribute this will fail.
}
```

Without `skip_type_params`, the following code is generated:

```rust
impl<T> ::scale_info::TypeInfo for A<T>
where
    T: Config,
    T: TypeInfo + 'static,      // <-- TypeInfo required for T
    T::Balance: TypeInfo + 'static,
{
    type Identity = Self;
        Type::builder()
        .type_params(vec![
            TypeParameter::new("T", Some(meta_type::<T>())), // <-- MetaType recorded for T
        ])
        /// .. elided
    }
};
```

With the `skip_type_params` applied 

```rust
impl<T> ::scale_info::TypeInfo for A<T>
where
    T: Config,
    T: 'static, // <-- No more `TypeInfo` bound
    T::Balance: TypeInfo + 'static,
{
    type Identity = Self;
        Type::builder()
        .type_params(vec![
            TypeParameter::new("T", None), // <-- MetaType not added for T
        ])
        /// .. elided
    }
};
```

Note how in the above we still get the autogenerated bounds for the associated type `T::A`

## Scenario 2: `skip_type_params` used in conjunction with `bounds`.

#88 adds support for specifying custom bounds, which overrides all auto generated bounds. Without `skip_type_params`, it is therefore required that the user manually adds `T: TypeInfo` to the custom bounds attribute.

```rust
#[derive(TypeInfo)]
#[scale_info(bounds(
  T: TypeInfo + 'static, // <-- The user needs to add this to make it compile
  U: TypeInfo + 'static,
))]
struct S<T, U> {
    marker: PhantomData<T>,
    field: U,
}
```

Adding `skip_type_params` means that we no longer need to add the unnecessary bound to our custom bounds.

```rust
#[derive(TypeInfo)]
#[scale_info(bounds(U: TypeInfo + 'static))]
#[scale_info(skip_type_params(T))]
struct S<T, U> {
    marker: PhantomData<T>,
    field: U,
}
```

### How `bounds` and `skip_type_params` interact.

When used independently, both attributes have an effect of the `where` clause predicate of the generated `TypeInfo` impl block.

- `bounds` replaces all autogenerated bounds with the user supplied ones in the attribute
- `skip_type_params` prevents the `TypeInfo` bound for the given type params being added to the autogenerated bounds.

When they are both used together, `skip_type_params` has no effect on the final generated `where` clause at all, because the predicates supplied in `bounds` are used verbatim instead of any autogenerated bounds.


test_suite/tests/derive.rs
```diff
@@ -609,6 +609,36 @@ fn doc_capture_works() {
...
+    #[scale_info(skip_type_params(T))]
+    struct Greet<T> {
+        marker: PhantomData<T>,
+    }
+
+    struct SomeType;
+
+    let ty = Type::builder()
+        .path(Path::new("Hey", "derive"))
+        .type_params(tuple_meta_type!(u16))
```
thiolliere, line 632: it seems to me more correct to me to have the first type param marked as `skipped` than ignored, otherwise when looking at the metadata and the doc of the rust type we can't easily tell which one is skipped and which one is kept

dvdplm: You mean like `.composite(Fields::named().field(|f| f.ty::<Greet<SomeType>>().name("ciao").type_name("Greet<T>").skipped_type_params::<T>())`?

thiolliere: actually I'm not sure anymore, I don't understand enough the usecase for type_params.

But what I meant is that currently the `Hey<SomeType, u16>` type information about type params would be `u16`, whereas I would have expected `"skipped", u16`.

derive/src/utils.rs
```diff
@@ -55,6 +59,41 @@ pub fn get_doc_literals(attrs: &[syn::Attribute]) -> Vec<syn::Lit> {
...
+            params,
+        })
+    }
+}
+
+syn::custom_keyword!(skip_type_params);
+
+/// Look for a `#[scale_info(skip_type_params(…))]`in the given attributes.
+///
+/// If found, do not register the given type params or require `TypeInfo` bounds for them.
```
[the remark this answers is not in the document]
ascjones: Outdated after refactor

test_suite/tests/derive.rs
```diff
@@ -609,6 +609,36 @@ fn doc_capture_works() {
...
 
+#[test]
+fn skip_type_params() {
+    #[allow(unused)]
+    #[derive(TypeInfo)]
+    #[scale_info(skip_type_params(T))]
+    struct Hey<T, N> {
+        ciao: Greet<T>,
+        ho: N,
+    }
```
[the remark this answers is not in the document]
ascjones: Yes it should be on the top level.

Robbepop:
Meeting idea/proposal:
```rust
#[derive(TypeInfo)]
#[scale_info(bounds(
    U: TypeInfo + 'static,
    T: skip,
))]
struct S<T, U> {
    marker: PhantomData<T>,
    field: U,
}
```

test_suite/tests/derive.rs
ascjones: This is a good point, we are losing some information by removing it entirely. I have addressed this by adding a `TypeParameter` struct with `Option<MetaType>` which would be `None` for skipped type params.

derive/src/trait_bounds.rs
```diff
@@ -39,12 +38,19 @@ use crate::{
...
-/// Ignores any type parameters not included in `type_params`.
+/// # Effect of attributes
+///
+/// `#[scale_info(skip_type_params(..))]`
+///
+/// Will not add `TypeInfo` bounds for any type parameters skipped via this attribute.
+///
+/// `#[scale_info(bounds(..))]`
+///
+/// Replaces *all* auto-generated trait bounds with the user-defined ones.
```
dvdplm, line 49:
```rust
#[scale_info(skip_type_params(U))]
#[scale_info(bounds("U: Other + Traits + NotTypeInfoStuff))]
struct A<T, U> { … }
```

In the above, I will not get `TypeInfo` bounds auto-generated for `T` (because `bounds()` override **all** auto generating machinery) but I will get the bounds I manually specified for `U` (because it's `bounds(…)` trumps `skipe_type_params`). Is that correct?

A few examples like above would be good! :)

src/lib.rs
```diff
@@ -101,6 +101,32 @@ macro_rules! tuple_meta_type {
     }
 }
 
+/// Construct a vector of `TypeParameter`s from pairs of the name and the concrete type.
```
dvdplm, line 104:
```suggestion
/// Construct a [`TypeParameter`] from the name and the concrete type.
```

ascjones:
> Meeting idea/proposal:
> Instead of this:
> 
> ```rust
> #[derive(TypeInfo)]
> #[scale_info(bounds(U: TypeInfo + 'static))]
> #[scale_info(skip_type_params(T))]
> struct S<T, U> {
>     marker: PhantomData<T>,
>     field: U,
> }
> ```
> 
> We could write this:
> 
> ```rust
> #[derive(TypeInfo)]
> #[scale_info(bounds(
>     U: TypeInfo + 'static,
>     T = skip,
> ))]
> struct S<T, U> {
>     marker: PhantomData<T>,
>     field: U,
> }
> ```

With `skip_type_params`, we want to keep all the other autogenerated bounds (see "Scenario 1: skip_type_params on its own"). In your example this is not a problem, but falls down when associated types are used in the fields of the type. We would have to manually add all of these, since `bounds` replaces all autogenerated bounds with the user supplied ones. 

This makes it unsuitable for the use case in substrate where we add `skip_type_params` to the proc-macro generated `Event`, `Call`, `Error` types. This solution would require adding back the bounds on any associated types.

> >     T = skip,

Also this would break the parsing of `syn::WherePredicates` which we use for extracting the bounds from the attribute.

src/lib.rs
ascjones: This macro actually does construct a vector, so it expands `named_type_params![(T, u16), (U, u32) ]` to `vec![ TypeParameter("T", MetaType::new::<u16>()), TypeParameter("U", MetaType::new::<u32>())]`

derive/src/trait_bounds.rs
ascjones: > In the above, I will not get TypeInfo bounds auto-generated for T (because bounds() override all auto generating machinery) but I will get the bounds I manually specified for U (because bounds(…) trumps skip_type_params). Is that correct?

Correct. It trumps it in respect to the generated `where` clause. However `skip_type_params` still has its effect on the generated list of `type_params` for the type.

> A few examples like above would be good! :)

I am in the process of adding more rustdocs on the topic, in the meantime I have added some more ui tests and have updated the PR description.

dvdplm (approved):

PR #97, opened 2021-06-16
Use wildcard for `compile_fail` tests

ascjones:
In order to automatically handle additions/removals of `fail_*` trybuild tests.

Also removes the now invalid `fail_non_static_lifetime.rs` (there is a `pass` equivalent)


test_suite/tests/derive.rs
```diff
@@ -613,9 +613,6 @@ fn doc_capture_works() {
 #[test]
 fn ui_tests() {
     let t = trybuild::TestCases::new();
-    t.compile_fail("tests/ui/fail_missing_derive.rs");
-    t.compile_fail("tests/ui/fail_unions.rs");
-    t.compile_fail("tests/ui/fail_use_codec_attrs_without_deriving_encode.rs");
-    t.compile_fail("tests/ui/fail_with_invalid_codec_attrs.rs");
+    t.compile_fail("tests/ui/fail_*.rs");
```
Robbepop, line 616: nice, I was not aware that this is possible ...

PR #98, opened 2021-06-18
Add BitVec support

ascjones:
Required for `polkadot`, some types use `BitVec` e.g. https://github.com/paritytech/polkadot/blob/master/primitives/src/v1/mod.rs#L468.

Rationale is that `parity-scale-codec` has support for bitvec: https://github.com/paritytech/parity-scale-codec/blob/master/src/bit_vec.rs, by enabling the `bit-vec` feature, so this library should too.

**EDIT**

I've modified this so that the `TypeDefBitSequence` struct and variant is still included even without `bit-vec` enabled, in order for compatible encoding/decoding between the crate with or without the feature enabled. Enabling the feature will simply add the `bitvec` dependency and the relevant`TypeInfo` impls.

### DIFF
--- a/src/ty/mod.rs
+++ b/src/ty/mod.rs
@@ -116,8 +116,14 @@ impl_from_type_def_for_type!(
     TypeDefTuple,
     TypeDefCompact,
     TypeDefPhantom,
-    TypeDefBitVec,
 );
+
+#[cfg(feature = "bit-vec")]
+impl From<crate::TypeDefBitVec> for Type {
+    fn from(item: crate::TypeDefBitVec) -> Self {
+        Self::new(Path::voldemort(), Vec::new(), item, Vec::new())
+    }
+}
 
 impl Type {
     /// Create a [`TypeBuilder`](`crate::build::TypeBuilder`) the public API for constructing a [`Type`]
@@ -198,6 +204,7 @@ pub enum TypeDef<T: Form = MetaForm> {
     Compact(TypeDefCompact<T>),
     /// A PhantomData type.
     Phantom(TypeDefPhantom<T>),
+    #[cfg(feature = "bit-vec")]
     /// A BitVec type.
     BitVec(TypeDefBitVec<T>),
 }
@@ -215,6 +222,7 @@ impl IntoPortable for TypeDef {
             TypeDef::Primitive(primitive) => primitive.into(),
             TypeDef::Compact(compact) => compact.into_portable(registry).into(),
             TypeDef::Phantom(phantom) => phantom.into_portable(registry).into(),
+            #[cfg(feature = "bit-vec")]
             TypeDef::BitVec(bitvec) => bitvec.into_portable(registry).into(),
         }
     }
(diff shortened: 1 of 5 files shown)


src/ty/mod.rs
```diff
@@ -485,3 +497,58 @@ where
         &self.type_param
     }
 }
+
+/// Type describing a [`bitvec::vec::BitVec`].
+#[cfg(feature = "bit-vec")]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(any(feature = "std", feature = "decode"), derive(scale::Decode))]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Debug)]
+pub struct TypeDefBitVec<T: Form = MetaForm> {
```
Robbepop, line 506: I'd rename this to `TypeDefBitSequence` to match with `Sequence` type definition.

Robbepop (approved):
I personally dislike the `bitvec` crate for being overcomplicated but if it is used pervasively by Polkadot/Sustrate and is thereby supported by `parity_scale_codec` we should also support it. Implementation looks fine to me.

src/ty/mod.rs
```diff
@@ -197,6 +204,9 @@ pub enum TypeDef<T: Form = MetaForm> {
     Compact(TypeDefCompact<T>),
     /// A PhantomData type.
     Phantom(TypeDefPhantom<T>),
+    #[cfg(feature = "bit-vec")]
+    /// A BitVec type.
+    BitVec(TypeDefBitVec<T>),
```
Robbepop, line 209: I think dtolnay mentioned that for compatibility with respect to different crate feature sets it is better to replace the inner type of `BitVec` with a placeholder type (dummy) of `not(feature = "bit-vec")` instead of entirely removing this variant.

src/ty/mod.rs
```diff
@@ -485,3 +497,58 @@ where
...
+        TypeDefBitVec {
+            bit_store_type: registry.register_type(&self.bit_store_type),
+            bit_order_type: registry.register_type(&self.bit_order_type),
+        }
+    }
+}
+
+#[cfg(feature = "bit-vec")]
+impl TypeDefBitVec {
+    /// Creates a new phantom type definition.
```
dvdplm, line 527:
```suggestion
    /// Creates a new `BitVec` type definition.
```

dvdplm (approved):
Maybe you could move the bitvec stuff into it's own module (and file) to cut down on the `#[cfg]` noise a bit?

Either way, lgtm.

src/ty/mod.rs
ascjones: That is much better, will avoid issues with incompatible decoding if we add variants after this.

ascjones: I've removed the feature flag from this variant now so it will be binary compatible, and also so that decoding/deserializing is supported even without the feature enabled.

Robbepop (approved):
LGTM

PR #99, opened 2021-06-21
Implement TypeInfo for VecDeque

ascjones:
Is encoded like a `Vec`: https://github.com/paritytech/parity-scale-codec/blob/master/src/codec.rs#L970

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -21,6 +21,7 @@ use crate::prelude::{
     collections::{
         BTreeMap,
         BTreeSet,
+        VecDeque,
     },
     marker::PhantomData,
     string::String,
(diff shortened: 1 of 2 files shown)


Robbepop (commented):
I cannot believe we forgot about this. Good catch!

Robbepop (approved):

ascjones:
> I cannot believe we forgot about this. Good catch!

Well, you can thank `rustc` for the catch.

src/impls.rs
```diff
@@ -126,6 +127,17 @@ where
     }
 }
 
+impl<T> TypeInfo for VecDeque<T>
+where
+    T: TypeInfo + 'static,
+{
+    type Identity = Vec<T>;
```
dvdplm, line 134: Why not `type Identity = [T];` here?

ascjones: You are right, it should be this in order that `[T]`, `Vec<T>` and `VecDeque<T>` all point to the same type

ascjones: https://github.com/paritytech/scale-info/pull/100

PR #101, opened 2021-06-23
Add `docs` feature

ascjones:
Docs will not be generated by the proc macro unless the `docs` feature is enabled. 

This allows reducing the size of the metadata code, useful for substrate runtimes: https://github.com/paritytech/substrate/discussions/8370#discussioncomment-909365

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ std = [
 derive = [
     "scale-info-derive"
 ]
-# Include code docs in type metadata.
+# Include rustdoc strings in the type metadata.
 docs = [
     "scale-info-derive/docs"
 ]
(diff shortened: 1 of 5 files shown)


derive/Cargo.toml
```diff
@@ -19,3 +19,8 @@ quote = "1.0"
 syn = { version = "1.0", features = ["derive", "visit", "visit-mut", "extra-traits"] }
 proc-macro2 = "1.0"
 proc-macro-crate = "1"
+
+[features]
+default = []
+# Include code docs in type metadata.
+docs = []
```
dvdplm, line 26: Wouldn't we want the docs to be included by default? In the steady-state future scenario we'll want the full metadata I think, and excluding it is something we want to do only while figuring out where to store it yes?

Cargo.toml
```diff
@@ -30,6 +30,10 @@ std = [
 derive = [
     "scale-info-derive"
 ]
+# Include code docs in type metadata.
```
dvdplm, line 33:
```suggestion
# Include rustdoc strings in the type metadata.
```

dvdplm (approved):
My only concern is whether we should include the docs feature in the defaults or not. I'd be inclined to include it by default.

PR #103, opened 2021-06-25
Impl TypeInfo for `NonZero*`

dvdplm:
Impl TypeInfo for NonZero*


Robbepop (approved):
I thought these impls already existed in scale-info. Urgently needed if that was not the case for some reason.

ascjones (approved):

PR #106, opened 2021-06-29
Reexport `parity-scale-codec` for derive 

### LINKED ISSUE #68: Derive requires direct dependency on `parity-scale-codec`
Discussed [here](https://github.com/paritytech/parity-common/pull/519#issuecomment-779863300) and [here](https://github.com/paritytech/scale-info/pull/61#discussion_r568760753).

Using the derive crate assumes a direct dependency on `parity-scale-codec` from the user's crate. Brainstorming ideas:

-  as [Basti suggests](https://github.com/paritytech/scale-info/pull/61#discussion_r568760753) do a reexport of the required types (see comment)
- better error message if the crate is not present
- add a feature `"compact"` or so to enable that functionality for which the codec types are required.
- make the macro determine whether any compact fields are present and only import the lib.

This also feeds into the larger discussion about how these two libraries coexist, and to what extent they should be coupled.

/cc @Robbepop @dvdplm @ordian

ascjones:
Fixes #68.

Rexports `parity-scale-codec` as `scale` for use in the derive macro.

@Robbepop's main issue with this was https://github.com/paritytech/scale-info/pull/61#discussion_r568764168:

> the scale-info crate works with multiple parity-scale-codec versions. So re-exporting one of the supported versions would case a lock-in and potentially a duplication of dependencies in case the scale-info user does not also use the same version.

However I believe we avoid this issue by having a suitably relaxed major version constraint on the `parity-scale-codec` dependency as per https://crates.io/crates/scale-info/0.7.0/dependencies.

image

Note that currently in the master branch we have restricted to a `codec` pre-release, but will likely loosen that back to major versions once `2.2.0` is released.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -65,7 +65,6 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
     let attrs = attr::Attributes::from_ast(&ast)?;
 
     let scale_info = crate_name_ident("scale-info")?;
-    let parity_scale_codec = crate_name_ident("parity-scale-codec")?;
 
     let ident = &ast.ident;
 
@@ -75,7 +74,6 @@ fn generate_type(input: TokenStream2) -> Result<TokenStream2> {
         &ast.generics,
         &ast.data,
         &scale_info,
-        &parity_scale_codec,
     )?;
 
     let (impl_generics, ty_generics, _) = ast.generics.split_for_impl();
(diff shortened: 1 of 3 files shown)


Robbepop:
> However I believe we avoid this issue by having a suitably relaxed major version constraint on the parity-scale-codec dependency as per https://crates.io/crates/scale-info/0.7.0/dependencies.

I support this and am convinced this is the right step forward for this crate.

Robbepop (approved):

PR #108, opened 2021-07-06
Option constructor macro hygiene

ascjones:
Fully qualify Option constructors to avoid collisions.


Robbepop (approved):
good catch!

ascjones:
> good catch!

Compiler caught it for me in polkadot :see_no_evil:

PR #110, opened 2021-07-09
Add index getter to `Variant`

ascjones:


src/ty/variant.rs
```diff
@@ -230,6 +230,11 @@ where
         &self.fields
     }
 
+    /// Returns the index of the variant, if specified.
+    pub fn index(&self) -> Option<u8> {
+        self.index
+    }
```
Robbepop, line 236: What is the difference here from the discriminant?

ascjones: The discriminant is only valid for "C Like" enums. 

If both are specified then the `index` takes precedence for encoding. We still retain the discriminant though for e.g. reconstructing the original type.

Robbepop: Ah okay I see. Although we should keep in mind that there is a Rust RFC to allow for discriminants for data variants. Although then `#[scale(index = N)]` would still override those which should be fine.

ascjones: Yes that would still hold true. You are correct that `index` specifically refers to the value from the attribute. And it can be different from the discriminator, but will take precedence for encoding.

src/ty/variant.rs
```diff
@@ -230,7 +230,7 @@ where
         &self.fields
     }
 
-    /// Returns the index of the variant.
+    /// Returns the index of the variant, if specified.
```
dvdplm, line 233: 👍

src/ty/variant.rs
dvdplm: > keep in mind that there is a Rust RFC

If/when that lands and is stabilised we could deprecate the `#[scale(index = N)]` attribute and only use discriminants. That would be nice (and a breaking change both to us and `parity-scale-codec`).

ascjones: Discriminant removed in https://github.com/paritytech/scale-info/pull/112

PR #111, opened 2021-07-12
Erase PhantomData fields

ascjones:
Feedback from @jacogr while working on https://github.com/polkadot-js/api/pull/3759.

He rightly points out that `PhantomData` fields are redundant from a scale encoding perspective, since nothing is actually encoded. So all type generators for different langs would have to implement a filter for them or have a bunch of useless fields. This is a `Rust` implementation detail leaking out into the metadata definition. This PR removes that burden and filters out fields of `PhantomData` at source.

I know there has been much discussion of this in the past, but I think this solution avoids the main problem we had in the past by avoiding the macro having to syntactically identify `PhantomData` types.

Let me know what you think @Robbepop @dvdplm

### DIFF
--- a/src/ty/composite.rs
+++ b/src/ty/composite.rs
@@ -89,9 +89,9 @@ impl IntoPortable for TypeDefComposite {
     }
 }
 
-impl TypeDefComposite {
+impl TypeDefComposit0e {
     /// Creates a new struct definition with named fields.
-    pub fn new<I>(fields: I) -> Self
+    pub(crate) fn new<I>(fields: I) -> Self
     where
         I: IntoIterator<Item = Field>,
     {
--- a/src/ty/composite.rs
+++ b/src/ty/composite.rs
@@ -89,7 +89,7 @@ impl IntoPortable for TypeDefComposite {
     }
 }
 
-impl TypeDefComposit0e {
+impl TypeDefComposite {
     /// Creates a new struct definition with named fields.
     pub(crate) fn new<I>(fields: I) -> Self
     where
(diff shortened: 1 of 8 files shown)


Robbepop:
Do I understand correctly that this PR filters out `PhantomData` types type safely?

ascjones:
> Do I understand correctly that this PR filters out `PhantomData` types type safely?

Precisely

src/ty/composite.rs
```diff
@@ -91,7 +91,7 @@ impl IntoPortable for TypeDefComposite {
 
 impl TypeDefComposite {
     /// Creates a new struct definition with named fields.
-    pub fn new<I>(fields: I) -> Self
+    pub(crate) fn new<I>(fields: I) -> Self
```
dvdplm, line 94: Why this change?

dvdplm (approved):
Full circle then! :)

@jacogr is right of course but [the reason we had `PhantomData`](https://github.com/paritytech/scale-info/pull/111/files#diff-f0a49acd88197542a00d5451a4541105af21f32e83a98e8b0f306f6e33040fa0L489-L498) here was that we decided to put the onus on the devs to `#[scale(skip)]` such types, so they should never appear in metadata (and thus not create issues for different languages) if we're disciplined. 

That said I don't disagree removing them.

src/ty/composite.rs
ascjones: So users can't instantiate with any fields containing a `PhantomData`. They are filtered out in the builder, and this would allow circumventing the builder. 

The other option of course would be to do the phantom filtering directly in this method, but I think we should encourage the use of the builder as the way to construct `Composite` types.

PR #112, opened 2021-07-15
Make variant index explicit, remove discriminant

ascjones:
Based on feedback from @jacogr.

Explicitly specify enum variant index. Prevents consumers having to duplicate the logic of inferring missing indices by the position in the list of variants.

Also removes `discriminant`, it was actually populated with the same value as `index` would be anyway. It's a legacy of when this was aiming to be general purpose Rust static reflection lib. Now we are a SCALE specific lib we should remove this Rust specific concept.


Robbepop (approved):
I agree we should not put Rust specifics into this format.

PR #115, opened 2021-07-16
Fix `PhantomData` filtering

ascjones:
Following up on #111, which failed to filter out `PhantomData` for tuple structs, causing a panic. I've removed said panic, replaced with a placeholder type def which can be searched for in case one sneaks through.


jacogr:
The good news is that I'm not seeing any "PhantomData placeholder" strings in the output now.

PR #120, opened 2021-07-29
Limit the amount of captured docs

ascjones:
Adds the `docs(max_paragraphs = N)` option to limit the capture docs as a mechanism for reducing the binary size.

Implemented despite @Robbepop having persuaded me that  we shouldn't do this, on the grounds that we should keep the API simple and not add too many bells and whistles. 

I wanted to see what effect on the final binary size [adding `max_paragraphs = 1`](https://github.com/paritytech/substrate/commit/847e1c1ec2c3daf631f44f391a073cdf66bf72e6) to the generated `Call` enum. The reason we can do this is that @jacogr displays only the first paragraph of those docs in the UI.

The answer is that it reduces the final compressed size by 32K, making the increase with the new metadata compared to `master` 68K instead of 100K. See https://github.com/paritytech/substrate/pull/8615#issuecomment-887281108
image

I'll leave this WIP, and if we decide that the tradeoff is worth it then we can can come back to it.

### DIFF
--- a/derive/src/attr.rs
+++ b/derive/src/attr.rs
@@ -166,8 +166,7 @@ impl Parse for ScaleInfoAttr {
             let docs = input.parse()?;
             Ok(Self::Docs(docs))
         } else {
-            Err(input
-                .error("Expected one of: `bounds`, `skip_type_params` or `docs"))
+            Err(input.error("Expected one of: `bounds`, `skip_type_params` or `docs"))
         }
     }
 }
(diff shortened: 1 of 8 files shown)


jacogr:
As discussed, I think it is worth-doing (especially in the case of Call) since subsequent paragraphs are really "how to Rust-dev-this-interface" comments in 99.999% of the cases.

It comes with a caveat - all extra code has a maintenance cost, so if that is not a nightmare, I think the value-add is there.

ascjones:
Doing some housekeeping, close this as stale. Maybe worth picking up again in the future but not something I plan to work on.

PR #122, opened 2021-08-03
Adds a type parameter getter

montekki:
Needed for ascjones/chameleon#5 to work

### DIFF
--- a/src/ty/mod.rs
+++ b/src/ty/mod.rs
@@ -215,6 +215,11 @@ where
     pub fn ty(&self) -> Option<&T::Type> {
         self.ty.as_ref()
     }
+
+    /// Get the name of the parameter.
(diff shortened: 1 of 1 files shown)


src/ty/mod.rs
```diff
@@ -208,6 +208,13 @@ where
     pub fn new(name: T::String, ty: Option<T::Type>) -> Self {
         Self { name, ty }
     }
+
+    /// Get the type of the parameter.
+    ///
+    /// `None` if the parameter is skipped.
+    pub fn ty(&self) -> Option<&T::Type> {
+        self.ty.as_ref()
+    }
```
ascjones, line 217: Could you add `name` too while you're here?

ascjones (approved):

Issue #123, opened 2021-08-14
Support for Range/RangeInclusive


jacogr:
These are now supported in scale-codec after https://github.com/paritytech/parity-scale-codec/pull/269

It may "just work", but worth a check.

dvdplm:
No, it'll need explicit support I think. I'll see if I can make it happen this week.

ascjones:
Note that this has not made it into a `parity-scale-codec` release yet, but we should be ready for when it does.

PR #124, opened 2021-08-23
Add support for Range and RangeInclusive

dvdplm:
Adding support for `Range` and `RangeInclusive` (added in https://github.com/paritytech/parity-scale-codec/pull/269 to `parity-scale-codec` and to polkadot.js in https://github.com/polkadot-js/api/pull/3791).

Closes https://github.com/paritytech/scale-info/issues/123

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -26,6 +26,7 @@ use crate::prelude::{
     marker::PhantomData,
     string::String,
     vec::Vec,
+    ops::Range,
 };
 
 use crate::{
@@ -38,6 +39,8 @@ use crate::{
     TypeDefPrimitive,
     TypeDefSequence,
     TypeDefTuple,
+    TypeDefRange,
+    TypeDefRangeInclusive,
     TypeInfo,
 };
 use core::num::{
@@ -52,6 +55,7 @@ use core::num::{
     NonZeroU64,
     NonZeroU8,
 };
+use std::ops::RangeInclusive;
 
 macro_rules! impl_metadata_for_primitives {
     ( $( $t:ty => $ident_kind:expr, )* ) => { $(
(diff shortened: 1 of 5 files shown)


src/impls.rs
```diff
@@ -347,6 +351,26 @@ where
...
+    }
+}
+
+impl<Idx> TypeInfo for RangeInclusive<Idx>
+where
+    Idx: TypeInfo + 'static + PartialOrd + core::fmt::Debug,
+{
+    type Identity = Self;
+    fn type_info() -> Type {
+        TypeDefRangeInclusive::new::<Idx>().into()
```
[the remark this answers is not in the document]
ascjones: Could add an `inclusive: bool` field? Then we would no longer need the two top level variants.

src/ty/mod.rs
```diff
@@ -411,6 +415,50 @@ where
...
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Debug)]
+pub struct TypeDefRange<T: Form = MetaForm> {
+    start: T::Type,
+    end: T::Type,
+    inclusive: bool,
+}
+
+impl TypeDefRange {
+    /// Creates a new [`TypeDefRange`] for the supplied index type. Use the `inclusive` parameter to
+    /// distinguish between [`Range`] and [`RangeInclusive`] range types.
```
ascjones, line 437: I don't think these links will resolve because `Range` and `RangeInclusive` are not in scope here?

src/impls.rs
```diff
@@ -347,6 +352,26 @@ where
     }
 }
 
+impl<Idx> TypeInfo for Range<Idx>
+where
+    Idx: TypeInfo + 'static + PartialOrd + core::fmt::Debug,
```
ascjones, line 357: We should be consistent and use `fmt::Debug` from the prelude, also for other occurrences of `core::fmt::Debug`

ascjones:
> Fixed the ui tests in #125

Merged this, so can update this branch now

src/ty/mod.rs
dvdplm: good pt!

test_suite/tests/json.rs
```diff
@@ -360,6 +364,27 @@ fn test_enum() {
...
+        closed_range: RangeInclusive<u64>,
+    }
+
+    assert_json_for_type::<TypeWithRanges>(json!({
+        "path": ["json", "TypeWithRanges"],
+        "def": {
+            "composite": {
+                "fields": [
+                    { "name": "open_range", "type": 0, "typeName": "Range<u8>" },
+                    { "name": "closed_range", "type": 2, "typeName": "RangeInclusive<u64>" },
```
ascjones, line 381: I think this test should also check that the `inclusive` field is set correctly on the respective types. Might be easiest by just comparing the whole serialized type registry which we do in some other tests.

ascjones (approved):
Apart from the nitpick about the test, LGTM!

test_suite/tests/json.rs
```diff
@@ -372,17 +372,56 @@ fn test_ranges() {
...
+                    "def": { "primitive": "u64" },
+                }
+            }
+        ],
+    });
+
+    let registry: PortableRegistry = registry.into();
+    assert_eq!(serde_json::to_value(registry).unwrap(), expected);
+
+    // assert_json_for_type::<TypeWithRanges>();
```
ascjones, line 424: remove commented out

PR #126, opened 2021-08-24
Add range getters, combine start and end types

ascjones:
Follow up to #124 

Getters are required for the range fields, since they are private.

Also I have combined the start and end fields because they are always the same type.


dvdplm (approved):
lgtm

PR #127, opened 2021-08-25
Release v0.11.0

ascjones:
### Added
- Add type parameter getters [(#122)](https://github.com/paritytech/scale-info/pull/122)
- Add support for Range and RangeInclusive [(#124)](https://github.com/paritytech/scale-info/pull/124), [(#126)](https://github.com/paritytech/scale-info/pull/126)
- Explicit codec indices for `TypeDef` and `TypeDefPrimitive` enums [(#127)](https://github.com/paritytech/scale-info/pull/127)

Also this moves the `Range` variant to the end, for backwards compatibility. /cc @jacogr

### DIFF
--- a/src/ty/mod.rs
+++ b/src/ty/mod.rs
@@ -247,14 +247,14 @@ pub enum TypeDef<T: Form = MetaForm> {
     Array(TypeDefArray<T>),
     /// A tuple type.
     Tuple(TypeDefTuple<T>),
-    /// A Range type.
-    Range(TypeDefRange<T>),
     /// A Rust primitive type.
     Primitive(TypeDefPrimitive),
     /// A type using the [`Compact`] encoding
     Compact(TypeDefCompact<T>),
     /// A type representing a sequence of bits.
     BitSequence(TypeDefBitSequence<T>),
+    /// A Range type.
+    Range(TypeDefRange<T>),
 }
 
 impl IntoPortable for TypeDef {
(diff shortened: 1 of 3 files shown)


src/ty/mod.rs
```diff
@@ -247,14 +247,14 @@ pub enum TypeDef<T: Form = MetaForm> {
...
-    /// A Range type.
-    Range(TypeDefRange<T>),
     /// A Rust primitive type.
     Primitive(TypeDefPrimitive),
     /// A type using the [`Compact`] encoding
     Compact(TypeDefCompact<T>),
     /// A type representing a sequence of bits.
     BitSequence(TypeDefBitSequence<T>),
+    /// A Range type.
+    Range(TypeDefRange<T>),
```
dvdplm, line 257: Oh good point. Do you mind mentioning this in the docs as well?

ascjones: That it is backwards compatible, or that new variants should be added to the end?

dvdplm: The latter. Something like "Tooling relying on scale-info data generally relies on order to find items. For this reason it is important that public enums add new variants items to the end to ensure backwards compatibility"?

jacogr: +1

ascjones: I've gone a step further and added explicit indices for this and `TypeDefPrimitive`, in case people delete or reorder variants.

dvdplm (approved):

PR #130, opened 2021-08-31
Replace Range variant with built-in composite definitions

ascjones:
This now describes precisely the encoding of the two range types introduces in https://github.com/paritytech/parity-scale-codec/pull/269.

The extra variant added in https://github.com/paritytech/scale-info/pull/124 would mean that if additional `Range` types were added it would have to be a breaking change. 

Also `Range` type definitions are not general SCALE concepts, they are just another "built-in" type that we can add to the prelude - e.g. `Option`, `Result` etc.

Required for #129 to prevent possible future breaking changes.

/cc @jacogr

### DIFF
--- a/src/ty/mod.rs
+++ b/src/ty/mod.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use crate::prelude::{
-    fmt,
     vec,
     vec::Vec,
 };
@@ -115,7 +114,6 @@ impl_from_type_def_for_type!(
     TypeDefArray,
     TypeDefSequence,
     TypeDefTuple,
-    TypeDefRange,
     TypeDefCompact,
     TypeDefBitSequence,
 );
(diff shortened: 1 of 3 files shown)


thiolliere:
This makes more sense to me, range is like any other composite type which contains inner types, the logic associated to its usage doesn't need to be in the primitive of scale-info but instead just a description similar to other composite types.

src/ty/mod.rs
```diff
@@ -271,9 +269,6 @@ pub enum TypeDef<T: Form = MetaForm> {
     /// A type representing a sequence of bits.
     #[codec(index = 7)]
     BitSequence(TypeDefBitSequence<T>),
-    /// A Range type.
-    #[codec(index = 8)]
-    Range(TypeDefRange<T>),
```
dvdplm, line 276: Do you think we should make `TypeDef` be `#[non_exhaustive]`?

dvdplm (approved):
Seems strictly better than my version.

jacogr:
My core needs a for “worse”, but since scale-info is not mapping to constructs, seems “sensible”, aka like Option would need to do more magic to detect.

src/ty/mod.rs
ascjones: I'm not sure because all `TypeDef` variants need to be handled usually, a wildcard might not make sense.

PR #139, opened 2022-01-21
Add #[must_use] to builder types

dvdplm:
Seems like the CI's Rust version's clippy has return-self-not-must-use enabled. Not sure I'm 100% convinced we need this but it makes some sense that builders be `#[must_use]` I guess.


ascjones (approved):
UI test needs fixing though

PR #140, opened 2022-02-06
Prepare release with SCALE 3.0

wigy-opensource-developer:


Cargo.toml
```diff
@@ -16,10 +16,10 @@ include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 [dependencies]
 bitvec = { version = "0.20.1", default-features = false, features = ["alloc"], optional = true }
 cfg-if = "1.0"
-scale-info-derive = { version = "1.0.0", path = "derive", default-features = false, optional = true }
+scale-info-derive = { version = "1.0.1", path = "derive", default-features = false, optional = true }
```
dvdplm, line 19: We'll need #137 too before releasing and I think that'll mean `1.1`? Wdyt @jsdw? @ascjones ?

dvdplm (approved):
LGTM. We'll need an entry in the `CHANGELOG` too though.

Cargo.toml
ascjones: We may have to make this `2.0` since the major bump of the requirement for `parity-scale-codec` makes it a breaking change.

wigy-opensource-developer: Scale 3 is also edition 2021 with an increased minimum Rust version, so I completely agree there.

dvdplm: I seem to be unable to formulate a rule about how to deal with dependency updates, how a breaking change downstream translates. It seems to be a common problem. Then there's the matter of MSRV, which I don't think we're taking into consideration at all here.

wigy-opensource-developer: There is no real consensus in the Rust community either. With a complex project like Substrate, I would err on too much safety and more manual upgrades, and implement some helper tools to make the overview of outdated deps easier.

dvdplm:
@wigy-opensource-developer #137 is merged now, so we can pull that in here and set the version to 2.0.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.0.0] - 2022-02-07
+
+## Changed
+- Upgraded to parity-scale-codec 3.0
+- Upgraded to bitvec 1.0
+- Minimum Rust version is 1.56.1 for editition 2021
```
jsdw, line 15: editition :D

jsdw (approved):

wigy-opensource-developer:
Jinx. Did a `git fetch -p` then `git push` and there was a commit in between. :smile:

jsdw:
Hehe, sorry!

PR #145, opened 2022-03-08
Add ability to reexport crate

### LINKED ISSUE #144: Feature request: ability to reexport crate
Hi everyone!

This is similar to https://github.com/paritytech/parity-scale-codec/issues/323 . The idea is to introduce the same attribute so the following will work:
```rust
use scale_info_reexport::info::TypeInfo;

#[derive(TypeInfo)]
#[scale_info(crate = scale_info_reexport::info)]
enum TestEnum {
    FirstVariant,
    SecondVariant,
}

```

gshep:
Closes #144

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scale-info"
-version = "2.0.1"
+version = "2.1.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2021"
 rust-version = "1.56.1"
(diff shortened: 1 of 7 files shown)


ascjones (approved):
Thanks!

jsdw (commented):
Awesome, looks great!

We should add a couple of UI tests for this (happy for this to be a separate piece of work); potentially we can rename `scale-info` in the test `cargo.toml` and then `use foo::{self as scale_info};` at the top of most tests to expose it in the "usual place", and then we can fully test that `#[scale_info(crate = "foo")]` works :)

jsdw (approved):

PR #147, opened 2022-03-14
Test the crate renaming

### LINKED ISSUE #146: Add UI tests for crate renaming
#145 adds the ability to tell scale-info where the crate lives, which is super handy in some circumstances.

let's add a couple of UI tests for this feature.

Potentially we can rename `scale-info` in the test `Cargo.toml` and then `use foo::{self as scale_info};` at the top of most tests to expose it in the "usual place", and then we can fully test that `#[scale_info(crate = "foo")]` works when it is only present under eg `foo::`.

jsdw:
Tweak the tests so that we can test crate renaming

I removed `rustversion`; it didn't appear to make a difference whether it was used or not any more. Can we move to 100% stable in CI too?

I also updated the copyright header and dates to 2022.

Closes #146


ascjones (approved):
LGTM

lexnv (approved):

dvdplm (commented):
nice.

gshep:
Should this version be published on crates.io?

PR #156, opened 2022-05-17
Strip invisible delimiters from type name

### LINKED ISSUE #155: Strip pretty-printed invisible delimiters for type names
See https://github.com/rust-lang/rust/issues/97076 and https://github.com/rust-lang/rust/pull/96682.

This can happen where `TypeInfo` is derived for a type generated by a `ty` expression in a declarative macro. This is parsed as a `syn::Type::Group`, which is then pretty-printed with the invisible delimiters `"/*»*/"` and `"/*«*/"`.

These should be stripped, possibly in `clean_type_string`, but maybe there is a cleaner way to do it.

ascjones:
Closes #155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.1.2] - 2022-05-18
+
+### Fixed
+- Strip invisible delimiters from type name [(#156)](https://github.com/paritytech/scale-info/pull/156)
+
 ## [2.1.1] - 2022-04-11
 
 ### Fixed
--- a/test_suite/tests/derive.rs
+++ b/test_suite/tests/derive.rs
@@ -870,8 +870,6 @@ fn ranges() {
 
 #[test]
 fn strips_invisible_delimiters_from_type_name() {
-    // When a type is specified as part of a macro, the tokens passed to the `TypeInfo` derive macro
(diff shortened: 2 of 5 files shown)


bkchr (approved):
Nice and even with a test :heart:

test_suite/tests/derive.rs
```diff
@@ -867,3 +867,24 @@ fn ranges() {
 
     assert_type!(Rangey, ty);
 }
+
+#[test]
+fn strips_invisible_delimiters_from_type_name() {
+    // When a type is specified as part of a macro, the tokens passed to the `TypeInfo` derive macro
+    // are a type `Group`, which is pretty printed with invisible delimiters e.g. /*«*/ bool /*»*/.
```
dvdplm, line 874: Maybe move/copy this comment into the code as well? Seems like a non-obvious gotcha?

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.1.2] - 2022-05-18
+
+### Fixed
+- Strip invisible delimiters from type name [(#156)](https://github.com/paritytech/scale-info/pull/156)
```
dvdplm, line 13: Should be under `unreleased` maybe?

dvdplm (approved):

CHANGELOG.md
ascjones: Yeah but I was planning on releasing straight from this commit to avoid a separate PR

lexnv (approved):

ascjones:
https://crates.io/crates/scale-info/2.1.2

PR #162, opened 2022-07-11
Fix cliff hanger in documentation

gilescope:
Feel free to tweak, but I was reading it and then I got left hanging...


dvdplm (approved):
ty

PR #163, opened 2022-07-16
additional support to allow frame-metadata to derive TypeInfo

gilescope:
I'm trying to see whether I can make frame-metadata have TypeInfos so that I can try decoding it in a no_std context. To do so it looks like these types need type info to get something that compiles.


jsdw:
Cool! Do we even want these behind a feature flag? It seemsl ike a generally useful and not super invasive change to make generally.

The only thing I don't like so much is the manual impl/use on `std::any::TypeId`, since it could change under our noses at any time. But I doubt that'll happen any time soon, and in the meantime perhaps we can migrate to our own version to avoid this.

PR #164, opened 2022-08-11
Construct `PortableRegistry` dynamically at runtime

ascjones:
Updates all type builders to allow construction of `PortableForm` types, and introduces the `PortableRegistryBuilder` to allow construction of a `PortableRegistry` dynamically. Currently this is only possible from an instance of the `Registry` which requires static type definitions.

The motivation for this is for `solang` to be able to generate metadata for `ink!` contracts, where static rust types are not available (since it is compiling Solidity contracts).

It is also generally useful to be able to dynamically define portable type hierarchies e.g. for post processing of statically generated metadata, or for testing.

**Note:** It is a goal that this should *not* be a breaking change and can be a minor release. So reviewers please look out for that.


src/registry.rs
```diff
@@ -53,11 +54,11 @@ pub trait IntoPortable {
     fn into_portable(self, registry: &mut Registry) -> Self::Output;
 }
 
-impl IntoPortable for &'static str {
+impl IntoPortable for <MetaForm as Form>::String {
```
xermicus, line 57: Can this be restricted to `String` only? This conflicts with all the other `IntoPortable` impls for the specs, but works fine when I change it to  `String`.

ascjones: I was able to remove this completely :point_down:

xermicus: That helps

src/meta_type.rs
```diff
@@ -96,6 +96,16 @@ impl MetaType {
         }
     }
 
+    /// Creates a new meta type from the user supplied type id and type info function.
+    ///
+    /// NOTE: It is the responsibility of the caller to ensure unique type ids per custom type.
+    pub fn new_custom(type_id: u64, fn_type_info: fn() -> Type<MetaForm>) -> Self {
```
xermicus, line 102: Can `fn_type_info` be opened up to accept parameters? Without the ability to inject any dependencies into this function, It is actually not of much use if we want to use this in a really dynamic fashion...

xermicus: See #166

ascjones:
Modified this to allow constructing of `PortableRegistry` and temporarily made type fields public to allow direct construction with `PortableForm`. See https://github.com/paritytech/cargo-contract/issues/659#issuecomment-1222443013

src/meta_type.rs
ascjones: I don't think this will work.

xermicus:
> Modified this to allow constructing of `PortableRegistry` and temporarily made type fields public to allow direct construction with `PortableForm`. See [paritytech/cargo-contract#659 (comment)](https://github.com/paritytech/cargo-contract/issues/659#issuecomment-1222443013)

Thanks! I couldn't get it to work (or figure a sane way to do so) neither with the static fn pointer. This looks promising now, I'll let you know!

xermicus (approved):
We have a working PoC in solang to re-use the `ink_metadata` crate for ABI generation, which relies currently on this branch. Thanks @ascjones

xermicus:
@ascjones I tried to make it a non-breaking change with a new `dynamic` feature flag. I get the current `master` of `ink!` pass the `--workspace` test when I use this branch now for all its crates. What do you think?

src/tests.rs
```diff
@@ -363,3 +364,56 @@ fn basic_enum_with_index() {
...
+
+#[test]
+fn construct_portable_registry() {
+    let mut types = Vec::new();
+
+    let u32_type_id = types.len() as u32;
+    types.push(PortableType::new_custom(
+        u32_type_id,
+        Type::new(Path::default(), vec![], TypeDefPrimitive::U32, vec![]),
+    ));
```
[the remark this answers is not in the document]
ascjones: As it stands, yes. Though this test was just part of a PoC. Ideally we could do this automatically with this guarantee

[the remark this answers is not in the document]
jsdw: Cool beans; yeah it'd be good to guarantee this via the API exposed :)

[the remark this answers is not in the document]
ascjones: Implemented in https://github.com/paritytech/scale-info/pull/164/commits/3c9d8dd4af746d97e37802fae9a66010dcc50f43

src/impls.rs
```diff
@@ -337,7 +337,7 @@ impl<T> TypeInfo for PhantomData<T> {
         // Fields of this type should be filtered out and never appear in the type graph.
         Type::builder()
             .path(Path::prelude("PhantomData"))
-            .docs(&["PhantomData placeholder, this type should be filtered out"])
+            .docs(["PhantomData placeholder, this type should be filtered out"])
```
ascjones, line 340: Note that this is not a required change, just a clippy suggestion since arrays implement `IntoIter`

src/ty/mod.rs
```diff
@@ -205,8 +218,14 @@ where
     T: Form,
 {
     /// Create a new [`TypeParameter`].
-    pub fn new(name: T::String, ty: Option<T::Type>) -> Self {
-        Self { name, ty }
+    pub fn new<S>(name: S, ty: Option<T::Type>) -> Self
+    where
+        S: Into<T::String>,
```
jsdw, line 223: This is very silly/edge casey but it's possible for this to be a breaking change because the type checker doesn't know quite so much about what `name` could be any more. Stupid contrived example:

```rust
fn into_string(val: impl Into<String>) {}
fn concrete_string(val: String) {}

fn main() {
    let maybe_string = ['a','b','c'].into_iter().collect();
    // breaks if changed to into_string:
    concrete_string(maybe_string);
}
```

src/ty/path.rs
```diff
@@ -186,19 +201,19 @@ mod tests {
     #[test]
     fn path_ok() {
         assert_eq!(
-            Path::from_segments(vec!["hello"]),
+            Path::<MetaForm>::from_segments(vec!["hello"]),
```
jsdw, line 204: I guess this highlights that making the functions no longer be only on `MetaForm`, code could break since we need the extra type information sometimes.

ascjones: I've moved these back to MetaForm now so should be non-breaking

jsdw (approved):
Looks good to me! Excited about the new `PortableRegistryBuilder`; I'm almos certain there will be places I can use that to simplify stuff.

As we chatted about, I think that while this technically could introduce breaking changes owing to loosening of types (and less type information for the compiler to work with), the acid test will be whether it fails to compile on some crates that use this stuff heavily like `substrate`, `scale-decode` and whatever else. So I'd be happy with this being a minor bump if it doesn't cause any breakage on this set of crates known to use it a bunch (a mini crater run if you will).

src/ty/mod.rs
ascjones: Possibly, but I have split this into `new` (legacy for MetaForm) and added a new `new_portable` method as I have for other types here. https://github.com/paritytech/scale-info/pull/164/commits/c9149c2e88812af32d6a85b9edcdc97c2f837652

ascjones:
>  as you said the acid test will be whether it fails to compile on some crates that use this stuff heavily like substrate, scale-decode and whatever else. So I'd be happy with this being a minor bump if it doesn't cause any breakage on this set of crates known to use it a bunch (a mini crater run if you will).

Indeed, as part of the release prep for `2.3` I will definitely check that all these things compile with no changes.

PR #166, opened 2022-08-22
switch fn_type_info to a FnMut

xermicus:
We should be able to inject dependencies into the `fn_type_info` function somehow.

### DIFF
--- a/src/meta_type.rs
+++ b/src/meta_type.rs
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::{
+    Arc,
+    Mutex,
+};
+
 use crate::prelude::{
     cmp::Ordering,
     fmt::{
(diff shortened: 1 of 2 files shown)


src/meta_type.rs
```diff
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::{
```
ascjones, line 15: Remember that this needs to work inside of Substrate runtimes which are `no_std`.

xermicus: A right. Then this won't fly. Not sure how we can pull this off then...

ascjones: Indeed it is a tricky problem, needs some thinking about...

xermicus: Maybe I can get this to work by implementing it as an n-ary function, I'm experiment RN

PR #168, opened 2022-09-12
Remove requirement on bitvec for std feature and constructing TypeDefBitSequence.

### LINKED ISSUE #167: Don't enable bitvec when "std" feature is enabled
Currently we have:

```toml
std = [
    "bitvec/std",
    "scale/std",
]
```

As the `std` feature in Cargo.toml. Enabling it enables `bitvec` too.

Let's swap to:

```toml
std = [
    "bitvec?/std",
    "scale/std",
]
```

To avoid enabling `bitvec` unless it's also asked for?

jsdw:
Requires Rust 1.60 for the optional dependency feature (? in Cargo.toml).

- Use the "?" feature in cargo.toml features to avoid bringing `bitvec` in when the "std" feature is enabled, unless it's been specifically asked for. I don't believe that this is a breaking change, since while the bitvec crate would be included if just the "std" feature was enabled and isn't now, it looks like we hide that stuff behind an explicit "bit-vec" feature anyway, so hopefully it just avoids pulling in a crate we don't actually use in that case.
- Allow `TypeDefBitSequence::new()` to be called without the `bit-vec` feature being enabled. This relaxes the constraints on it slightly, making "incorrect" uses more possible, but allowing us to create the type info without expecting any specific underlying library. Again, since this just relaxes/adds, it's not a breaking change.

I also updated the UI test output just because it had changed a touch. There are a couple of unused lifetime warnings that can't be silenced but will be silencable in a future rust update (see https://github.com/rust-lang/rust/issues/96956)

parity-scale-codec PR to help with this: https://github.com/paritytech/parity-scale-codec/pull/375

Closes #167

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,9 +23,9 @@ derive_more = { version = "0.99.1", default-features = false, features = ["from"
 scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive"] }
 
 [features]
-default = ["std"]
+default = ["std", "bit-vec"]
 std = [
-    "bitvec/std",
+    "bitvec?/std",
     "scale/std",
 ]
 derive = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ derive_more = { version = "0.99.1", default-features = false, features = ["from"
 scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive"] }
 
 [features]
-default = ["std", "bit-vec"]
+default = ["std"]
 std = [
     "bitvec?/std",
     "scale/std",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "scale-info"
 version = "2.1.2"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2021"
-rust-version = "1.56.1"
+rust-version = "1.60.0"
 
 license = "Apache-2.0"
 readme = "README.md"
--- a/src/ty/mod.rs
+++ b/src/ty/mod.rs
@@ -576,17 +576,21 @@ where
     }
 }
 
-#[cfg(feature = "bit-vec")]
 impl TypeDefBitSequence {
     /// Creates a new [`TypeDefBitSequence`] for the supplied bit order and bit store types.
-    pub fn new<T, O>() -> Self
+    ///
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ scale = { package = "parity-scale-codec", version = "3", default-features = fals
 [features]
 default = ["std"]
 std = [
-    "bitvec/std",
+    "bitvec?/std",
```
jsdw, line 28: The bitvec crate will no longer be pulled in when "std" is enabled, but this I think makes no difference to the external interface since bitvec additions are behind the separate "bit-vec" feature anyway.

src/ty/mod.rs
```diff
@@ -576,17 +576,21 @@ where
...
 
-#[cfg(feature = "bit-vec")]
 impl TypeDefBitSequence {
     /// Creates a new [`TypeDefBitSequence`] for the supplied bit order and bit store types.
-    pub fn new<T, O>() -> Self
+    ///
+    /// With the `bit-vec` feature enabled, the expected usage is to provide either
+    /// `bitvec::order::Lsb0` or `bitvec::order::Msb0` as the order type, and then something
+    /// like u8, u8, or u32 as the store type. Without the `bit-vec` feature enabled, it's
+    /// recommended that your types have identical `TypeInfo` to those.
```
dvdplm, line 585: …or else? :)
Like, is "recommended" the right word here? Isn't it more "has to be"?

dvdplm (approved):

src/ty/mod.rs
jsdw: I mean, in theory it can be whatever it wants; it's just that our current tooling won't know how to decode the thing if the types don't line up with what we expect (no reason somebody else can't make tooling that can support more exotic types though if they want to go that route)

ascjones (approved):

lexnv (approved):

Cargo.toml
jsdw: Ooh ok; I'll do a quick follow PR there then to prevent that :)

jsdw: PR for parity-scale-codec thing: https://github.com/paritytech/parity-scale-codec/pull/375

jsdw:
I'll merge this for now but ideally would make sure we bump the parity-scale-codec version (if needed) when that PR lands before releasing.

PR #171, opened 2022-11-14
Change NonZero* TypeInfo implementation to not be recursive

benluelo:
The current `TypeInfo` implementation for the `NonZero*` types is recursive, defining the types as containing a field of itself. This PR fixes the implementation to properly define the inner field as the `NonZero*`'s wrapped type.

I also considered if the `NonZero*` types should be encoded as their wrapped type (i.e. `<NonZeroU64 as TypeInfo>::Identity == u64`), but I'm not sure what implications that would have.

NOTE: I've forked this off of b881bfb94f0df02781c8f523fb96f02ea4ef13bf since our current lockfile contains scale-info = 2.2.


ascjones (approved):
LGTM thanks!

PR #174, opened 2023-03-08
portable: Retain the provided type IDs

lexnv:
This PR implements the `retain` method on the `PortableRegistry`.

The `retain` method allows tools (such as subxt or capi) to interpret the
RuntimeMetadata for only a subset of pallets.

The method retains only the portable types needed to express the provided ids.
A given type ID can be defined by nesting type IDs, such as the case
of a [`TypeDef::Composite`] and others. To retain a valid [`PortableRegistry`]
all the types needed to express an ID are included. Therefore, the number of
elements defined by the result equals or exceeds the number of
provided IDs.

This is implemented in two steps:
- recursively visit and collect all type IDs needed to express the given list of IDs
  - visiting happens in DFS manner
  - while visiting, type IDs are normalized (assigned a new incremental ID starting from zero)
- modify the nested type IDs to point to the new location of the type in the registry

### Testing Done
- unit tests
- tested in subxt with PoC branch: https://github.com/paritytech/subxt/tree/lexnv/pallet_metadata_poc

The following represents the metadata size for only one pallet, based on the PoC branch.
Further work is required to minimize the `runtime type` present in the Metadata to
reduce the size even more than the PoC. 
```bash
232K Vesting.scale
235K VoterList.scale
232K Whitelist.scale
451K polkadot_metadata.scale
````

A more realistic number lies between the two type ID dumps from https://github.com/paritytech/subxt/issues/629#issuecomment-1456761827.  

This is part of: https://github.com/paritytech/subxt/issues/629

// CC: @paritytech/tools-team

### CODE REVIEW

File: src/portable.rs [hunk 1]
```diff
@@ -152,6 +306,105 @@ impl PortableRegistryBuilder {
     }
 }
 
+/// Recursive resolver for the type IDs needed to express a given type ID.
+struct TypeIdResolver<'a> {
+    registry: &'a PortableRegistry,
+    result: BTreeMap<u32, u32>,
+    next_id: AtomicU32,
```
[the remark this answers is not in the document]
lexnv: I've used `AtomicU32` here having a compact way to express `compare_and_exchange` via `fetch_add`. Will write that manually

lexnv:
While at it, I've also tried to use the PortableRegistry to build out types without manually updating the nested type ids. 

### Registry from `retain_recursive_ids` test
```
0: u32
1: u64
2: Sequence { ty id 0}
3: MyStruct { primitive: 0, vec_of_u32: 2}
4: MyStructSecond { vec_of_u32: 2, second: 3}
```

### Using the `PortableRegistryBuilder` without updating nested type IDs
```
0: MyStructSecond { vec_of_u32: 2, second: 3}
1: Sequence { ty id 0 }
2: u32
3: MyStruct { primitive: 0, vec_of_u32: id 2}
```
I'm afraid that's not right because `MyStruct.primitive` should be u32 with ID 2, not ID 0.

### Using the current PR
```
0: MyStructSecond {vec_of_u32: 1, second: 3}
1: Sequence {ty id 2}
2: u32
3: MyStruct { primitive: 2, vec_of_u32: 1 }
```

src/portable.rs
```diff
@@ -70,6 +72,157 @@ impl PortableRegistry {
...
+    /// # Note
+    ///
+    /// A given type ID can be defined by nesting type IDs, such as the case
+    /// of a [`TypeDef::Composite`] and others. To retain a valid [`PortableRegistry`]
+    /// all the types needed to express an ID are included. Therefore, the number of
+    /// elements defined by the result equals or exceeds the number of provided IDs.
+    pub fn retain(
+        &mut self,
+        ids: impl IntoIterator<Item = u32>,
+    ) -> Result<Vec<u32>, PortableRegistryError> {
```
jsdw, line 95: Thinking about what `Vec::retain` etc looks like and the general usage, I was wondering whether a signature more like this would also work:

```rust
fn retain<F, R>(
    &mut self,
    filter: F,
    on_retained: R
) 
where
    F: FnMut(u32) -> bool,
    R: FnMut(u32, u32)
{ /*...*/ }
```

Where `F` is called for each type we're not sure about keeping (some types will be kept by virtue of being needed recursively by other types, after all) to ask the user whether to keep it or not. `R` is called for any type that's kept (including recursive ones) to note the old and new ID.

The implementation could then do one pass over the "old" registry, pushing types (and recursively pushing subtypes that are needed) to it as it goes.

This would remove the need for any return type (and no more PortableRegistryError, since the user doesn't get to choose what type IDs to look for; only which to filter). It'd also allow the user to collect the "old -> new" Id mappings via the `R` function however they please (ie push to a vec as you have done, or to a map or do some update in place or whatever).

But that all said, in reality I can see that the current approach would work just fine for what we need too (ie, we'd collect all types mentioned in the frame-metadata types and pass them into here, obtaining a mapping we can then use to go back over said frame-metadata stuff to do the updating). Just sortof pondering how I'd approach it I guess :)

src/ty/fields.rs
```diff
@@ -78,22 +78,22 @@ pub struct Field<T: Form = MetaForm> {
...
     )]
-    type_name: Option<T::String>,
+    pub type_name: Option<T::String>,
     /// Documentation
     #[cfg_attr(
         feature = "serde",
         serde(skip_serializing_if = "Vec::is_empty", default)
     )]
-    docs: Vec<T::String>,
+    pub docs: Vec<T::String>,
```
jsdw, line 96: These being `pub` makes the impl's (`name()`. `type_name()` etc) sortof pointless. We'd have to decide here whether we are happy to expose this internal structure. Maybe `pub(crate)` for now if it's useful to avoid allocating when retaining, but it looks like you just `clone()` each type anyway, so perthaps these aren't needed at all? Or perhaps functions like `id_mut()` etc to go along with the existing non-mut ones is a way to go?

jsdw: Or maybe the way to go is to just clone each field and construct a new thing with `::new()` in each case? Not sure; I can see a loaf od stuff has been made `pub` in the file, and I thnk to avoid exposing a lot of internal details maybe just `pub(crate)` for now is easiest? (and you could update each type in-place and `mem::swap` or something to avoid allocations, maintaining a set of types you've visited so you never look at a type more than once anyway)

src/portable.rs
lexnv: That makes total sense and would boost the flexibility of the API. That's implemented now :D Thanks!

src/ty/fields.rs
lexnv: I've picked `pub(crate)` for the path of least resistance without needing to add `mut` counterparts for the getter functions or remove them entirely. Also, I've implemented a dummy `Type::Default` which assumes the default type is a `bool` to avoid extra allocations, thanks!

src/portable.rs
```diff
@@ -70,13 +73,135 @@ impl PortableRegistry {
...
+    /// # Note
+    ///
+    /// A given type ID can be defined by nesting type IDs, such as the case
+    /// of a [`TypeDef::Composite`] and others. To retain a valid [`PortableRegistry`]
+    /// all the types needed to express an ID are included. Therefore, the number of
+    /// elements defined by the result equals or exceeds the number of provided IDs.
+    pub fn retain<F, R>(&mut self, mut filter: F, mut on_retained: R)
+    where
+        F: FnMut(&u32) -> bool,
+        R: FnMut(&u32, &u32),
```
jsdw, line 96: We may as well just dereference since they are small copy types :)

src/ty/mod.rs
```diff
@@ -68,22 +68,38 @@ pub struct Type<T: Form = MetaForm> {
...
     #[cfg_attr(
         feature = "serde",
         serde(skip_serializing_if = "Vec::is_empty", default)
     )]
-    docs: Vec<T::String>,
+    pub(crate) docs: Vec<T::String>,
+}
+
+impl<T> Default for Type<T>
+where
```
ascjones, line 90: It would be nice not to have to implement `Default` here where it doesn't make sense. I see that it is required for `mem::take`, I wonder if it is possible to utilize `Option<Type>` in that case

src/ty/fields.rs
ascjones: The other option of course would have be to implement methods directly on the types, in a similar manner to the `InfoPortable` trait implementations. OTOH it may be a good thing to keep those implementations inside `retain`.

Taking the `IntoPortable` comparison a bit further, imagine a trait `RetainTypes`, implemented in a similar way, and then you would have to manually implement it in the same way for the metadata definitions. Then you could have an API like `PortableRegistry::retain_for<T: RetainTypes>(metadata_type: &T)`, obviously would still need to be able to provide multiple types somehow, but I think you get the drift.

src/portable.rs
```diff
@@ -70,6 +75,158 @@ impl PortableRegistry {
...
+            new_types: &mut Vec<PortableType>,
+            retained_mappings: &mut BTreeMap<u32, u32>,
+        ) -> u32 {
+            // Type already retained; just return the new ID for it:
+            if let Some(id) = retained_mappings.get(&id) {
+                return *id
+            }
+
+            // Zero-allocation default implementation that assumes
+            // the type def to be `TypeDefPrimitive::Bool`.
```
jsdw, line 109:
```suggestion
            // Zero-allocation default implementation that is used as
            // a placeholder and never accessed. 
```

jsdw (approved):
Looks great :)

src/ty/mod.rs
lexnv: I've picked the placeholder approach to avoid changing the types: `Vec<Port>` to `Option<Port>` and avoid an extra unwrap :D Let me know what you think of it Thanks!

src/portable.rs
```diff
@@ -70,6 +75,158 @@ impl PortableRegistry {
     pub fn types(&self) -> &[PortableType] {
         &self.types
     }
+
+    /// Retains only the portable types needed to express the provided ids.
+    ///
+    /// The type IDs retained are returned as key to the `HashMap`.
```
ascjones, line 81:
```suggestion
    /// The type IDs retained are returned as key to the [`BTreeMap`].
```

ascjones (approved):

PR #175, opened 2023-03-23
Release 2.4.0

lexnv:
### Added
- portable: Retain the provided type IDs [(#174)](https://github.com/paritytech/scale-info/pull/174)

// CC @paritytech/tools-team


jsdw:
I think 2.4 makes more sense given semver; crates depending on eg "2.3" should still automatically be happy to use this :)

jsdw:
Ooh also, it feels like the derive crate is generally kept in line (perhaps that was kissed with the latest patch release), so I wonder whether we should bump that to the same? @ascjones any opinions?

ascjones (approved):

niklasad1 (approved):

PR #176, opened 2023-03-28
ty: Make type fields public 

lexnv:
Subxt needs access to the types stored in the `PortableRegistry` to create a slim subset of the metadata,
where the generic types of a `Type` are ignored.

To achieve this, subxt will access something similar to:  `registry.types.get(index).ty.type_param`.

While at it, make the other fields public.  The old getters are still left in place for backward compatibility.

cc @paritytech/subxt-team

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -303,7 +303,7 @@ impl<F: Form, T> FieldsBuilder<F, T> {
 impl<T> FieldsBuilder<MetaForm, T> {
     fn push_field(mut self, field: Field) -> Self {
         // filter out fields of PhantomData
-        if !field.ty().is_phantom() {
+        if !field.ty.is_phantom() {
             self.fields.push(field);
         }
         self
--- a/src/interner.rs
+++ b/src/interner.rs
@@ -49,13 +49,17 @@ use serde::{
 pub struct UntrackedSymbol<T> {
     /// The index to the symbol in the interner table.
     #[codec(compact)]
-    id: u32,
+    pub id: u32,
     #[cfg_attr(feature = "serde", serde(skip))]
     marker: PhantomData<fn() -> T>,
 }
 
 impl<T> UntrackedSymbol<T> {
     /// Returns the index to the symbol in the interner table.
+    #[deprecated(
+        since = "2.5.0",
+        note = "Prefer to access the fields directly; this getter will be removed in the next major version"
+    )]
     pub fn id(&self) -> u32 {
         self.id
     }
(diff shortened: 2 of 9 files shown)


jsdw:
I could get behind making fields public and treating them as just bundles of data; a description of the shape of a type.

I wonder whether there are any forseeable reasons to keep this internal structure hidden; @ascjones perhaps you have thoughts?

If we do go this route I'd also be tempted to add a `#[deprecated(...)]` attr to each of the getter functions to encourage people over to just accessing the fields directly (it would be nice ultimately to be consistent and remove the getters in the next major version).

The attr could look like:

```rust
#[deprecated(since = "version.number", note = "prefer to access the fields directly; this getter will be removed in the next major version")]
```

ascjones:
I'm fine with making these public. 

Regarding the getters, probably best to deprecate them for the least disruption, then remove for a `3.0` release.

ascjones (approved):

jsdw:
This looks good to me!

I _think_ that `#[deprecated]` warnings in sub-dependencies won't trickle up (eg if A depends on B which depends on C, and C uses `#[deprecated]` on some stuff B is using, then if I compile A I would hope _not_ to see those warnings). I bring this up just because it could be quite noise for anybody transitively depending on scale-info anywhere if this isn't actually the case. Would be good to confirm if possible :)

jsdw:
I had a go locally and created 3 crates in 3 folders seen below. When I run`cargo check` in `temp` I see the deprecated warning. Maybe this is different for published crates, or maybe deprecated warnings do bubble up through all crates? 

### temp:

cargo.toml:

```
[dependencies]
temp2 = { path = "../temp2" }
```

lib.rs

```rust
pub fn add(left: usize, right: usize) -> usize {
    temp2::add(left, right)
}
```

### temp2

cargo.toml:

```
[dependencies]
temp3 = { path = "../temp3" }
```

lib.rs

```rust
pub fn add(left: usize, right: usize) -> usize {
    temp3::add(left, right)
}
```

### temp3

lib.rs

```rust
#[deprecated]
pub fn add(left: usize, right: usize) -> usize {
    left + right
}
```

lexnv:
I've had a go at this in https://github.com/paritytech/subxt/pull/879.

What I did there was to point scale-info to this branch in `Cargo.toml`

```bash
[patch.crates-io]
scale-info = { git = "https://github.com/paritytech/scale-info.git", branch = "lexnv/resolve_mut" }
```

Warnings did appear for all subxt usages across the repo, but after fixing those individually in [this commit](https://github.com/paritytech/subxt/pull/879/commits/e28bb8a3bebc1de4a66ba31e9e3655b3da3d7aff) they went away.

jsdw:
I asked in the reddit and warnings apparently only show up for local crates, not those published to crates.io, so I think all is good :)

jsdw (approved):

PR #178, opened 2023-04-22
Schema generation

SkymanOne:
This PR adds support for schema generation required for ink! metadata verification. All types are now derive `schemars::JsonSchema` under an optional feature flag `schema`

### DIFF
--- a/src/form.rs
+++ b/src/form.rs
@@ -39,16 +39,16 @@ use crate::{
     meta_type::MetaType,
 };
 
-#[cfg(feature = "schema")]
+#[cfg(all(feature = "std", feature = "schema"))]
 use schemars::JsonSchema;
 #[cfg(feature = "serde")]
 use serde::Serialize;
 
 /// Trait to support derivation of `JsonSchema` for schema generation.
-#[cfg(feature = "schema")]
+#[cfg(all(feature = "std", feature = "schema"))]
 pub trait JsonSchemaMaybe: JsonSchema {}
 /// Trait to support derivation of `JsonSchema` for schema generation.
-#[cfg(not(feature = "schema"))]
+#[cfg(not(all(feature = "std", feature = "schema")))]
 pub trait JsonSchemaMaybe {}
 
 /// Trait to control the internal structures of type definitions.
@@ -74,7 +74,7 @@ pub trait Form {
 ///
 /// Allows to be converted into other forms such as portable form
 /// through the registry and `IntoPortable`.
-#[cfg_attr(feature = "schema", derive(JsonSchema))]
+#[cfg_attr(all(feature = "std", feature = "schema"), derive(JsonSchema))]
 #[cfg_attr(feature = "serde", derive(Serialize))]
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug)]
 pub enum MetaForm {}
@@ -91,7 +91,7 @@ impl Form for MetaForm {
 /// This resolves some lifetime issues with self-referential structs (such as
 /// the registry itself) but can no longer be used to resolve to the original
 /// underlying data.
-#[cfg_attr(feature = "schema", derive(JsonSchema))]
+#[cfg_attr(all(feature = "std", feature = "schema"), derive(JsonSchema))]
 #[cfg_attr(feature = "serde", derive(Serialize))]
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug)]
 pub enum PortableForm {}
@@ -111,7 +111,8 @@ cfg_if::cfg_if! {
     }
 }
 
-// Blanket implementation
-impl JsonSchemaMaybe for &'static str {}
-#[cfg(any(feature = "std", feature = "decode"))]
-impl JsonSchemaMaybe for crate::prelude::string::String {}
+// Blanket implementations
+#[cfg(not(all(feature = "std", feature = "schema")))]
+impl<T> JsonSchemaMaybe for T {}
+#[cfg(all(feature = "std", feature = "schema"))]
+impl<T> JsonSchemaMaybe for T where T: JsonSchema {}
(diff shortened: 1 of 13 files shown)


src/form.rs
```diff
@@ -39,6 +39,9 @@ use crate::{
     meta_type::MetaType,
 };
 
+use cfg_if::cfg_if;
+#[cfg(all(feature = "std", feature = "schema"))]
```
ascjones, line 43: Does `schemars` require std?

src/meta_type.rs
```diff
@@ -111,3 +114,25 @@ impl MetaType {
         self == &MetaType::new::<crate::impls::PhantomIdentity>()
     }
 }
+
+#[cfg(all(feature = "std", feature = "schema"))]
+impl schemars::JsonSchema for MetaType {
```
ascjones, line 119: What do we need this for?

src/form.rs
SkymanOne: yep, it depends on `serde` internally and other `std` crates

ascjones: If it depends on `serde` then that will just be brought in automatically?

> and other std crates

What do you mean?

Anyway just wondering what the errors are if you remove the `feature = "std"`. Usually for this type of thing you should be able to configure the features in the manifest.

SkymanOne: > What do you mean?

It relies on `std` library

SkymanOne: I initially had it without `feature = "std"` and it errored out that some types can not be found in scope

SkymanOne: Actually tested without `feature = "std"` and it worked. Idk maybe some odd typo

src/form.rs
```diff
@@ -74,21 +87,54 @@ impl Form for MetaForm {
...
         impl Form for PortableForm {
-            type Type = UntrackedSymbol<TypeId>;
+            type Type = UntrackedSymbol<TypeIdDef>;
             // Owned string required for decoding/deserialization
             type String = crate::prelude::string::String;
         }
     } else {
         impl Form for PortableForm {
-            type Type = UntrackedSymbol<TypeId>;
+            type Type = UntrackedSymbol<TypeIdDef>;
```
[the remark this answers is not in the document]
SkymanOne: I avoided it, but `TypeId` is used in other places too like `Interner` and `Registry`

src/form.rs
ascjones: You still need to enable `std` to the `schema` feature (in `Cargo.toml`. I have just tried with `cargo check --no-default-features --features schema` and that fails because of https://github.com/paritytech/scale-info/blob/a69ac5516ab292176a51860ab8c87f89b49ff588/src/lib.rs#L15

src/meta_type.rs
ascjones: Indeed, I was hoping that the `derive(JsonSchema)` was clever enough to add associated trait bounds e.g. `where T::Type: JsonSchema. Sadly not though so we do need this.

src/form.rs
```diff
@@ -92,3 +110,8 @@ cfg_if::cfg_if! {
         }
     }
 }
+
+// Blanket implementation
+impl JsonSchemaMaybe for &'static str {}
+#[cfg(any(feature = "std", feature = "decode"))]
+impl JsonSchemaMaybe for crate::prelude::string::String {}
```
ascjones, line 117: True blanket impl would be `impl<T> JsonSchemaMaybe for T` for the feature not enabled and  `impl<T> JsonSchemaMaybe for T where T: JsonSchema` when the `schema` feature is enabled

ascjones (commented):
Please add a line here https://github.com/paritytech/scale-info/blob/ce21de5ef09c413161293422738172887e9ec639/.github/workflows/rust.yml#L45 to test that feature on its own e.g. `cargo check --no-default-features --features schema`. At the moment that fails for me.

src/form.rs
```diff
@@ -39,25 +39,42 @@ use crate::{
     meta_type::MetaType,
 };
 
+#[cfg(all(feature = "std", feature = "schema"))]
+use schemars::JsonSchema;
 #[cfg(feature = "serde")]
 use serde::Serialize;
 
+/// Trait to support derivation of `JsonSchema` for schema generation.
+#[cfg(all(feature = "std", feature = "schema"))]
```
SkymanOne, line 48: Turns out that `schemars` indeed can only be used with `std`. Trying to compile it with no-std throws errors like:
```
error[E0433]: failed to resolve: use of undeclared crate or module `std`
   --> src/ty/mod.rs:657:63
    |
657 | #[cfg_attr(all(feature = "serde", feature = "schema"), derive(schemars::JsonSchema))]
    |                                                               ^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `std`
    |
    = note: this error originates in the derive macro `schemars::JsonSchema` (in Nightly builds, run with -Z macro-backtrace for more info)
```

ascjones: Indeed it does. Did you try adding `"std"` into the `schema` feature in the manifest?

SkymanOne: Nope. I thought it is better that a developer explicitly enables `std` feature.

ascjones: Why? `schema` depends on `std` being enabled.

SkymanOne: okay, I don't have a strong opinion on that. I'll change it now

src/registry.rs
```diff
@@ -23,10 +23,11 @@
 //! namespaces. The normal Rust namespace of a type is used, except for the Rust
 //! prelude types that live in the so-called root namespace which is empty.
 
+use core::any::TypeId;
```
ascjones, line 26: Should use the prelude for consistency

ascjones (approved):
LGTM!

Should be good to be a minor (non-breaking) release :+1:

xermicus (approved):

PR #180, opened 2023-05-15
Implement `TypeInfo` for `Arc`

koute:
I'd like to replace a `Box` in one of the structs which implements `TypeInfo` with an `Arc`, but `TypeInfo` is not implemented for `Arc`, so let's make it that it is.

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -30,6 +30,7 @@ use crate::prelude::{
         RangeInclusive,
     },
     string::String,
+    sync::Arc,
     vec::Vec,
 };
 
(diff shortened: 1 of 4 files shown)


jsdw (approved):

niklasad1 (approved):

src/impls.rs
```diff
@@ -279,6 +280,17 @@ where
...
 }
 
+impl<T> TypeInfo for Arc<T>
+where
+    T: TypeInfo + ?Sized + 'static,
+{
+    type Identity = T;
+
+    fn type_info() -> Type {
+        Self::Identity::type_info()
```
ascjones, line 290: We may want to revisit the idea of adding some extra information about transparent wrapper types like this, e.g. for `Box` to avoid https://github.com/paritytech/subxt/blob/c08eb6c8f1ae1ea61964196733008cc3a4bfb627/codegen/src/types/composite_def.rs#L307

ascjones (approved):

PR #185, opened 2023-06-16
feat: Implement Rc TypeInfo ✨

EvolveArt:
I have a merkle tree implementation that uses Rc and I want to store that within my substrate runtime but it's not possible so let make it happen!

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -29,6 +29,7 @@ use crate::prelude::{
         Range,
         RangeInclusive,
     },
+    rc::Rc,
     string::String,
     sync::Arc,
     vec::Vec,
@@ -270,6 +271,17 @@ where
 }
 
 impl<T> TypeInfo for Box<T>
+where
+    T: TypeInfo + ?Sized + 'static,
+{
+    type Identity = T;
+
+    fn type_info() -> Type {
+        Self::Identity::type_info()
+    }
+}
+
+impl<T> TypeInfo for Rc<T>
 where
     T: TypeInfo + ?Sized + 'static,
 {
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -39,6 +39,7 @@ cfg_if! {
             string,
             sync,
             vec,
+            rc
         };
     } else {
         pub use alloc::{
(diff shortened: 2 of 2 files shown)


jsdw (approved):

niklasad1 (approved):

EvolveArt:
hey @niklasad1 want me to bump the version before merging ?

EvolveArt:
Clippy seems to be complaining for no reasons too

jsdw:
Cargo lock is meaningless in library crates (upstream dependencies will ignore it and generate their own cargo locks instead etc)

The clippy errors are possibly just new errors that appeared since rust was updated. The cleanest approach would be another PR that runs a clippy --fix etc to sort these out :)

EvolveArt:
Sounds good thanks for the quick answer @jsdw !

jsdw:
Oh and no need for any version bumps; that'll be taken care of when we do a release :)

EvolveArt:
> Oh and no need for any version bumps; that'll be taken care of when we do a release :)

Thanks I do have something weird happening in my project tho :
When using my fork as a dependency it breaks all the other scale-info types implementations :(

You can checkout this [PR](https://github.com/keep-starknet-strange/madara/pull/667) to see

jsdw:
> > Oh and no need for any version bumps; that'll be taken care of when we do a release :)
> 
> Thanks I do have something weird happening in my project tho : When using my fork as a dependency it breaks all the other scale-info types implementations :(
> 
> You can checkout this [PR](https://github.com/keep-starknet-strange/madara/pull/667) to see

I haven't checked, but I'm guessing it's because there are now multiple versions of scale-info in your cargo tree; use the cargo.toml patch stuff if you aren't already to ensure all scale-infos use your fork :)

EvolveArt:
Great! Could we merge this ? @jsdw or anything blocking let me know

jsdw:
> The clippy errors are possibly just new errors that appeared since rust was updated. The cleanest approach would be another PR that runs a clippy --fix etc to sort these out :)

This still needs fixing; I'll open a new PR to address those and then when this branch is updated and passing we can merge it!

jsdw:
Edit: nevermind; re-ran the job and it passed!

PR #192, opened 2023-08-15
Use default generic param

boundless-forest:
A small improvement: There  is already a default generic param, so don't need to add the type param explicitly.

```rs
pub struct Type<T: Form = MetaForm>  {
    ...
}
```


ascjones:
CI fails because of clippy errors see https://github.com/paritytech/scale-info/pull/193

ascjones (approved):
LGTM!

Once https://github.com/paritytech/scale-info/pull/193 is merged it should fix the CI here

niklasad1 (approved):

ascjones:
Merge with master, conflicts are just due to `cargo fmt` now being done with the `stable` toolchain

PR #195, opened 2023-08-18
Update test cases

boundless-forest:
The current test file looks a bit messy and some type cases are missing.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -17,15 +17,23 @@ use crate::{
     prelude::{
         borrow::Cow,
         boxed::Box,
-        collections::{BTreeMap, BTreeSet},
+        collections::{BTreeMap, BTreeSet, VecDeque},
+        ops::{Range, RangeInclusive},
+        rc::Rc,
         string::String,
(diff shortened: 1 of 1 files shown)


src/tests.rs
```diff
@@ -42,18 +50,103 @@ macro_rules! assert_type {
...
+    assert_type!(i128, TypeDefPrimitive::I128);
+    assert_type!(u8, TypeDefPrimitive::U8);
+    assert_type!(u16, TypeDefPrimitive::U16);
+    assert_type!(u32, TypeDefPrimitive::U32);
+    assert_type!(u64, TypeDefPrimitive::U64);
+    assert_type!(u128, TypeDefPrimitive::U128);
 
-    assert_type!([bool], TypeDefSequence::new(meta_type::<bool>()));
+    assert_type!(
+        NonZeroI8,
```
ascjones, line 68: Nitpick: the non zero types are not primitives. I would put them in their own test.

boundless-forest: Fixed.

PR #197, opened 2023-10-19
Implement `replace_segment` attribute

bkchr:
This implements the `replace_segment` attribute. This attribute can be used to rename one or multiple segments returned by `TypeInfo::path`. It could be for example be used to rename a changed crate name. For more info see the examples and the docs in this pr.


franciscoaguirre (approved):

KiChjang (approved):

derive/src/lib.rs
```diff
@@ -101,12 +101,23 @@ impl TypeInfoImpl {
...
+        });
+
         Ok(quote! {
             impl #impl_generics #scale_info::TypeInfo for #ident #ty_generics #where_clause {
                 type Identity = Self;
                 fn type_info() -> #scale_info::Type {
                     #scale_info::Type::builder()
-                        .path(#scale_info::Path::new(::core::stringify!(#ident), ::core::module_path!()))
+                        .path(#scale_info::Path::new_with_replace(
+                            ::core::stringify!(#ident),
```
ggwpez, line 117: Why not pass the ident as `syn::Ident`? I think it would be cleaner than calling `.split("::")` subsequently.

ggwpez (approved):
Just one nit.

derive/src/lib.rs
bkchr: We call `split` on the module path that comes out of the `module_path!` macro. Or I maybe misunderstand you?

acatangiu (approved):

lexnv (approved):
LGTM! 

Small nit: we'd also need an entry in the [changelog](https://github.com/paritytech/scale-info/blob/master/CHANGELOG.md) if we plan to release from this PR 🙏

derive/src/lib.rs
ggwpez: Oh, yea :man_facepalming:

PR #199, opened 2024-01-16
Introduce more solid `no_std` testing

tadeohepperle:
The `no_std` tests we had before, did not enforce that our dependencies are all `no_std`. E.g. any dependency could use the standard library and still our `no_std` tests would succeed.

We now circumvent this by defining a custom panic handler and a global allocator such that any dependencies that make use of the standard library will fail to compile, because in that case there would be more than one global allocator and panic handler. Inspiration was taken from [serde's no_std testing](https://github.com/serde-rs/serde/blob/master/test_suite/no_std/src/main.rs) (Thanks for the hint @niklasad1).

Testing in `no_std` is a bit difficult, but one approach I take here is just running a main function, that returns `Aborted (core dumped)` with a non-zero exit code if a panic is encountered (e.g. in an untrue assert). This can run in the CI to make sure the code runs and not only compiles in a `no_std` environment.

### DIFF
--- a/test_suite/derive_tests_no_std/Cargo.toml
+++ b/test_suite/derive_tests_no_std/Cargo.toml
@@ -10,7 +10,9 @@ license = "Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-scale-info = { path = "../..", default-features = false, features = ["derive"] }
+scale-info = { path = "../..", default-features = false, features = ["derive", "bit-vec", "decode"] }
+scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive", "bit-vec"] }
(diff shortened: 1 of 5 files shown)


test_suite/derive_tests_no_std/Cargo.toml
```diff
@@ -11,8 +11,13 @@ license = "Apache-2.0"
 
 [dependencies]
 scale-info = { path = "../..", default-features = false, features = ["derive"] }
-scale = { package = "parity-scale-codec", version = "2", features = ["full"] }
```
niklasad1, line 14: does it work with ` scale = { package = "parity-scale-codec", version = "2", default-features = false }`?

Or why was it removed?

tadeohepperle: I just removed it from the dependencies because it was not used before and is not needed.

lexnv (approved):

PR #204, opened 2024-03-22
Change type id to new type id in `PortableRegistry::retain()`.

tadeohepperle:
The type's id should be set to the new id to reflect the correct position in the `Vec` holding the types of the `PortableRegistry`.


jsdw (commented):
One small thing; this was missed in our retain tests before, so could we tweak the tests to make sure this works too? :)

jsdw (approved):

lexnv (approved):

PR #206, opened 2024-04-04
Tweak retain function to avoid issues with recursive types

jsdw:
An issue was spotted when metadata was being trimmed in Subxt that was traced back to ths `.retain()` function not working properly with recursive types (like `RuntimeCall`, which contains itself in eg the sudo call).

Here, we just tweak the order of things a little to avoid such an issue. I also re-did the retain related tests and added one to test for this issue (which fails until the fix here is applied). The original issue in https://github.com/paritytech/subxt/issues/1509 also seems to be resolved with this fix applied.

One of the interesting side effects of this change is that, previously, all of the IDs in a type would tend to be smaller than the ID of the type itself (because they would be added to the registry first), but after calling `.retain()`, the reverse will be true because we save a spot for the "parent" type before recursing and adding any child types. This shouldn't matter in practise as we shouldn't rely on the ordering of type IDs anyway.

I also bumped the version to 2.11.2 so that we can release a patch with this fix in when merged.


src/portable.rs
```diff
@@ -303,304 +304,390 @@ impl PortableRegistryBuilder {
...
-        let u32_type_id = builder.register_type(u32_type);
-        assert_eq!(U32_TY_ID, u32_type_id);
+    fn ty<T: TypeInfo + 'static>() -> MetaType {
+        MetaType::new::<T>()
+    }
 
-        let u64_type = Type::new(Path::default(), vec![], TypeDefPrimitive::U64, vec![]);
-        let u64_type_id = builder.register_type(u64_type);
-        assert_eq!(U64_TY_ID, u64_type_id);
+    fn make_registry(tys: impl IntoIterator<Item=MetaType>) -> (Vec<u32>, PortableRegistry) {
```
niklasad1:
```suggestion
    fn make_registry(tys: impl IntoIterator<Item = MetaType>) -> (Vec<u32>, PortableRegistry) {
```

src/portable.rs
```diff
@@ -292,304 +304,423 @@ impl PortableRegistryBuilder {
 
 #[cfg(test)]
 mod tests {
+    use scale::Compact;
+
     use super::*;
+    use crate::ty::TypeDefPrimitive;
     use crate::{build::*, prelude::vec, *};
 
-    // Type IDs generated by `build_registry`.
```
niklasad1, line 298: yeah ok these are not needed because make_registry

niklasad1 (approved):
nice

src/portable.rs
jsdw: Yeah exactly!

lexnv (approved):
Nice!

PR #208, opened 2024-05-08
Version 2.11.3

tdimitrov:
Bump the version of the crate to `2.11.3` and add `Cargo.lock` under version control.


ascjones (approved):

.gitignore
niklasad1, line 1: No, this is wrong `scale-info` is just a library and Cargo.lock doesn't matter.

Thus, revert this please

jsdw (approved):

.gitignore
tdimitrov: I just had exactly the opposite conversation with @ascjones.

Bottom line - I'll revert Cargo.lock and keep the comment. Okay @ascjones ?

jsdw: Nowadays the argument is that there are pros and cons either way.

Basically, the pros of keeping it are that one knows the precise deps the thing was tested against and can workout which dep updates may have broken something (if they do). The cons are that packages pulling in `scale-info` will ignore the lockfile, so you want to make sure not to be only testing what's in the lockfile in CI.

So if we start committing the lockfile, I'd prob suggest adding a `cargo update` to CI to make sure the latest versions of things work.

PR #209, opened 2024-07-19
Allow renaming struct fields

conr2d:
This PR introduces an attribute for struct fields to rename the name of a field.

```rs
#[derive(TypeInfo)]
struct S {
  #[scale_info(rename = "type")]
  r#type: bool,
}
```

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -184,12 +184,12 @@ impl TypeInfoImpl {
                 } else {
                     quote!(ty)
                 };
-                let name = if let Some(ident) = utils::maybe_renamed(f) {
-                    quote!(.name(#ident))
-                } else if let Some(ident) = ident {
-                    quote!(.name(::core::stringify!(#ident)))
-                } else {
-                    quote!()
+                let name = match utils::maybe_renamed(f) {
+                    Some(ident) => quote!(.name(#ident)),
+                    None => ident
+                        .as_ref()
+                        .map(|ident| quote!(.name(::core::stringify!(#ident))))
+                        .unwrap_or(quote!()),
                 };
                 quote!(
                     .field(|f| f
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -29,7 +29,7 @@ use syn::{
     punctuated::Punctuated,
     token::Comma,
     visit_mut::VisitMut,
-    Data, DataEnum, DataStruct, DeriveInput, Field, Fields, Ident, Lifetime,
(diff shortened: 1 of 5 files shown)


derive/src/utils.rs
```diff
@@ -99,7 +99,7 @@ where
     find_meta_item("codec", itr, pred)
 }
 
-fn find_meta_item<'a, F, R, I, M>(kind: &str, mut itr: I, mut pred: F) -> Option<R>
+pub(crate) fn find_meta_item<'a, F, R, I, M>(kind: &str, mut itr: I, mut pred: F) -> Option<R>
```
jsdw, line 102: Nit: rather than making this `pub(crate)`, perhaps expose another util like the above which looks for `#[scale_info(rename = "...")]` and returns something like `Option<&str>` if a rename happened or `None` if not?

derive/src/lib.rs
```diff
@@ -184,7 +184,9 @@ impl TypeInfoImpl {
                 } else {
                     quote!(ty)
                 };
-                let name = if let Some(ident) = ident {
+                let name = if let Some(ident) = utils::maybe_renamed(f) {
```
pkhry, line 187:
```suggestion
               match utils::maybe_renamed(f) {
                    Some(ident)  => quote!(.name(#ident)),
                     None  => {
                        ident.map(|ident| {
                            quote!(.name(::core::stringify!(#ident)))
                        }).unwrap_or(quote!())
                    },
                }                
```

pkhry (changes requested):

pkhry (approved):

test_suite/tests/derive.rs
```diff
@@ -907,3 +907,19 @@ fn replace_segments_works() {
...
     assert_type!(R, ty);
 }
+
+#[test]
+fn rename_field() {
+    #[allow(unused)]
+    #[derive(TypeInfo)]
+    pub struct S {
+        #[scale_info(rename = "type")]
+        r#type: bool,
```
jsdw, line 917: I'd like to see a more "dramatic" rename also in the test, eg "foo" renamed to "bar" (eg as a second field in this struct) and then a non-renamed field in the struct just to show that also ok in conjunciton with renamed ones

jsdw (approved):
LGTM!

If we merge this, we should also honour the attribtue in `scale-encode` and `scale-decode` so that deriving `EncodeAsType` / `DecodeAsType` on structs with renamed fields is in line with the derived typeinfo.

jsdw:
Merging latest master into this should fix the build error!

PR #220, opened 2025-02-19
fix: no-std derive start fn

### LINKED ISSUE #219: ci: fix broken `no_std`  job
The `start attribute` has been removed and should be replaced with `pub extern "C" fn _start` instead
For context see https://github.com/paritytech/scale-info/pull/218#issuecomment-2669007875

Failing job log here: https://github.com/paritytech/scale-info/actions/runs/13417793067/job/37482816223

saiintbrisson:
Closes #219 by switching to the manual `_start` function. It is required to compile the binary targeting a none-linux triple to avoid having it codegen the original `_start` function, which then requires `__libc_start_main`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+.PHONY: check test clean
+
+check:
+	cargo clippy --all-targets -- -D warnings
+	cargo clippy --all --all-features -- -D warnings
+	cargo check --no-default-features --features bit-vec
+	cargo check --no-default-features --features docs
+	cargo check --no-default-features --features serde
+	cargo check --no-default-features --features serde,decode
+	cargo check --no-default-features --features schema
+
+test:
+	cargo test --all --all-features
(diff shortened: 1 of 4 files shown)


Makefile
niklasad1, line 1: This makefile doesn't seem to be used? Can we remove it?

test_suite/derive_tests_no_std/src/main.rs
```diff
@@ -12,14 +12,23 @@
...
 
-#[start]
-fn start(_argc: isize, _argv: *const *const u8) -> isize {
+#[no_mangle]
+pub extern "C" fn _start() {
     test();
-    0
+
+    use core::arch::asm;
+    unsafe {
```
niklasad1, line 24: Can you add a comment that this inline assembly makes a `syscall exit(0)`?

niklasad1 (approved):
Lovely, thanks for fixing this

